PR #1, opened 2021-02-15
Proposal: IPFS Gateway HTTP API

anorth:
This is a project proposal for the web3 dev team.

### DIFF
--- a/proposals/IpfsGatewayHttpApi.md
+++ b/proposals/IpfsGatewayHttpApi.md
@@ -35,6 +35,7 @@ _What must be true for this project to matter?_
 - go/js-ipfs is slower and more difficult to interact with than an HTTP API
 - Dapp developers are satisfied with the imperfect decentralization of using a gateway as an intermediate step.
 - Having the same, more powerful HTTP API on localhost removes obstacles towards a deeper interaction with vendors like Brave
+- Low-powered IoT devices can leverage gateways as less expensive way of moving data around, with true p2p kicking in only in LAN or as a fallback when centralized gateways are down.
 
 #### User workflow example
 _How would a developer or user use this new capability?_
--- a/proposals/IpfsGatewayHttpApi.md
+++ b/proposals/IpfsGatewayHttpApi.md
@@ -89,6 +89,7 @@ _What specific deliverables should completed to consider this project done?_
 Our gateway serves arbitrary DAGs over simple HTTP requests. 
 Web-apps can fetch, manipulate, and render IPLD data with the standard JS XHR or [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) without the need for any client library. 
 Backends that embed/extend the gateway also support publishing and pinning arbitrary IPLD data.
+Built-in mechanism for access control exist and makes it easy to limit POST/PUT/pinning capability to specific origins.
 
 ####  What does success look like?
 _Success means impact. How will we know we did the right thing?_
--- a/proposals/IpfsGatewayHttpApi.md
+++ b/proposals/IpfsGatewayHttpApi.md
@@ -1,0 +1,134 @@
+# IPFS Gateway HTTP API
+Authors: @anorth
+Initial PR: TBD
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+
+Today, web and other clients must embed a full IPFS node implementation in order to interact with IPFS 
+on any level deeper than fetching flattened UnixFS file data from a gateway. 
+This embedding is heavyweight, and complex. 
+Maintaining a fully-featured js-ipfs implementation, as the only integration option, is a lot of work. 
+Delegating to a backend IPFS node directly using the RPC-style API is high friction.
+
+After this project, an HTTP Gateway API supports direct interaction with IPLD data using standard web/REST semantics, 
+but abstracts away the complexity of native protocol interactions (DHT, bitswap, graphsync) from clients. 
+A gateway client can fetch and verify arbitrary IPLD DAGs. 
+The API is application-level (DAG-at-a-time rather than node-at-a-time), maximising ease and minimising round-trips. 
+Scripts can fetch DAGs with curl to feed into other tools.
+
+Authenticated extensions to the API can push/pin IPLD data to the gateway for hosting. 
+The PL-operated gateway is read-only, but a Dapp provider can host their own for Dapp clients to interact with, or embed/extend it with application logic.
+
+"No API is the best API".
+Related ideas are in [Verifiable HTTP Gateway Responses](https://github.com/ipfs/in-web-browsers/issues/128), 
+[CAR export from HTTP Gateways](https://github.com/ipfs/in-web-browsers/issues/170).
+
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+
+- Dapp developers want to use IPFS, and there are lots of them
+- Dapps and other applications need a more complex data model than UnixFS files
+- Dapps and other integrations would find HTTP easier to interact with than low-level protocols
+- Embedding a full go/js-ipfs node is a significant barrier to client use
+- go/js-ipfs is slower and more difficult to interact with than an HTTP API
+- Dapp developers are satisfied with the imperfect decentralization of using a gateway as an intermediate step.
+- Having the same, more powerful HTTP API on localhost removes obstacles towards a deeper interaction with vendors like Brave
+
+#### User workflow example
+_How would a developer or user use this new capability?_
+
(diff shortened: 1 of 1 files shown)


anorth:
This is the first proposal, being used as a process example as well as a real proposal. So we're figuring out the workflow as we go.

I think review here is only about whether this is an appropriate proposal, and for discussion/collaboration on the proposal itself. Approval signals nothing about whether we should actually undertake the proposal at any point in time. We'll discuss that in separate issues.

anorth:
FYI @gmasgras @achingbrain @autonome

proposals/IpfsGatewayHttpApi.md
```diff
@@ -0,0 +1,134 @@
+# IPFS Gateway HTTP API
+Authors: @anorth
+Initial PR: TBD
```
willscott, line 3: Is this a self link to the PR that the file is created in, or initial development work for the proposal? If it's the former, it should be reasonably easy to look at history / git blame to find that initial conversation.

proposals/IpfsGatewayHttpApi.md
```diff
@@ -0,0 +1,134 @@
...
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+- Just add more RPCs for DAG traversal to go-ipfs, that Infura nodes can then serve. Expose these through JS and other client libraries.
+- Our gateways provide the read-only go-ipfs RPC API to the public, and establish that as standard
+- Amazing technical work so that embedding js-ipfs is both easier and more performant than using an HTTP API
+
+#### Dependencies/prerequisites
+
+- JS-IPLD libraries capable of core encodings and IPLD representation
+- IPLD selectors (for non-trivial DAG traversal)
```
willscott, line 117: just the golang / server side implementation, right?

warpfork:
:+1: :+1: 

Designing an API that's usable by a lightweight client, while also being designed explicitly to avoid "N+1 query problems", seems very high-value to me.

proposals/IpfsGatewayHttpApi.md
anorth: This PR. I've inserted the link now the PR exists.

proposals/IpfsGatewayHttpApi.md
anorth: JS implementation will be necessary for sub-DAG queries from the client, which would be needed for complex cases. So no, not just server side. But also we needn't block creating and exposing this API on that JS selector implementation (I gather it doesn't exist?). It will just be more useful once JS clients can express queries.

achingbrain:
I'm +1 for a simple API that doesn't need a complicated client to access.  However this proposal or the design of the outcome of this proposal should take into account the limitations of the existing HTTP API when used in anything that implements browser [Fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API).

TLDR: you cannot do full-duplex streaming, so design the API as simple request/response pairs, or only require streaming from the server to the client.

More detail:

1. Limited streaming support
    - We use multipart requests to simulate streaming from the client to the server
    - Fetch uses [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) which requires adding all parts to the request before sending - so no streaming uploads.
      - You can use [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob)s as parts, but these only work in a memory-efficient way if they are [File](https://developer.mozilla.org/en-US/docs/Web/API/File)s.
      - You can create a Blob from an array or a string, so you can have JSON Blobs (for example), but these are kept in memory until the FormData is disposed of, so it doesn't scale and you still need to add them all to the request before any of it is sent
      - According to the specs you can use a ReadbleStream as the request body, but it's not implemented anywhere (yet, I think Chrome just did an 'intent to ship').  Even if it is, there's no guarantee it'll be full-duplex (see next point)
    - Browsers require sending the entire request before a single byte from the response is made available to the calling code, so no full-duplex streaming
      - You can almost use `XMLHttpRequest` for this by accessing the `.responseText` property before the response completes, but the response body is buffered in-memory as text until the `.readyState` is `DONE`, so it's not a solution either
2. No HTTP trailers
    - Errors cannot be sent in HTTP trailers, they are not readable by fetch, instead they need to be sent in the response body

In js-IPFS we settled on using gRPC over websockets which addresses these points, though the API becomes harder to use without a client which runs counter to the spirit of this proposal.

This may not be an issue, most people reach for some sort of client when dealing with an API in order to get something that respects the conventions and norms of their development platform and to not have to deal with the weird foibles of whatever the API developers [thought was REST](https://martinfowler.com/articles/richardsonMaturityModel.html)* - the people cURLing APIs are not (I believe) the majority of Web2-graduating-to-Web3 developers, not after the first few minutes anyway.

Further reading:

* https://github.com/ipfs/js-ipfs/issues/2519
* https://github.com/ipfs/js-ipfs/pull/3403


----

* I'm kidding, no-one seems to know (or at least can agree on) what REST is

lidel (commented):
The true scope of this proposal will depend on how capable [Writable Gateway approach](https://discuss.ipfs.io/t/writeable-http-gateways/210?u=lidel) (added it as prior art below) turns out to be when it is not limited to files, but can operate on IPLD DAGs, control pinning, and perhaps support [resumable uploads](https://tus.io/).

But yes, I imagine it being simple request/response pairs with potential for leveraging https://streams.spec.whatwg.org when those features land in browsers.

We have distinct Gateway  (`/ipfs`, `/ipns`) and API (`/api/v0`)  endpoints today, and only the most basic read operations  can be executed via Gateway. The goal here would be to shift what is possible on the Gateway, so most popular operations work without touching advanced API.

Error reporting on `/api/v0` is broken in browsers, as noted by @achingbrain  in https://github.com/ipfs/js-ipfs/issues/2519,
so for everything not handled by Gateway (eg. pubsub, advanced streaming), we need to replace HTTP-based `/api/v0`  with something like RPC over websockets anyway (js-ipfs already does this: https://github.com/ipfs/js-ipfs/pull/3403).

proposals/IpfsGatewayHttpApi.md
```diff
@@ -0,0 +1,134 @@
+# IPFS Gateway HTTP API
```
lidel, line 1: Perhaps this would be less vague?

```suggestion
# Writable HTTP Gateway
```

anorth: Writeable is only part of it. I actually think the first 70% of value will come from readable/queryable DAGs

anorth:
Thanks! I've noted the HTTP streaming limitations in the body. @lidel I have also declared you a co-author, thanks for fleshing some bits out.

achingbrain:
> Error reporting on /api/v0 is broken in browsers

It's worse than that, errors and streaming are broken anywhere that uses Fetch - that's browsers, [node-fetch](https://www.npmjs.com/package/node-fetch), [electron-fetch](https://www.npmjs.com/package/electron-fetch), [React Native](https://reactnative.dev/docs/network), etc.

The only place it does work is via node's [http](https://nodejs.org/api/http.html) module, but Fetch is the thing that users reach for when they need to make a network request on pretty much every platform JS runs on and it will be what they use if you're going for clientless here.

Stebalien:
I'd like to avoid supporting _two_ APIs if possible unless we intend to deprecate/remove one of them. I'm concerned that either:

1. This limited REST API will eventually grow to feature parity with an RPC API.
2. Applications will inevitably outgrow the REST API and need to either use both APIs or migrate to the RPC API.

That's why I'm personally in-favor of the gRPC+websocket API.

On the other hand, I have occasionally wanted a slightly more featureful gateway, mostly for integration with scripts where there is no convenient js-ipfs-http-client. But I'd still prioritize making the more general purpose API suck less first.

lidel:
l feel we have a consensus that we need to deprecate `/api/v0` over HTTP due to reasons mentioned in  https://github.com/protocol/web3-dev-team/pull/1#issuecomment-780357792 and replace it with gRPC+websocket API as drop-in replacement that finally works like real RPC.  And that is a higher priority than making Gateway more useful.

Perhaps that should be a separate Proposal tho? 
I really feel  there is a merit in making Gateways more useful, and would like to keep this one as-is.

mikeal:
Love it.

One technical consideration I’d like to add: whenever we create an API or a protocol that uses selectors we should modularize the use of selectors so that, in the future, we could swap selectors out for another query mechanism and reuse the rest of the protocol or API. This should be easy, we just need to use a union and it shouldn’t add any extra burden to the implementation and will futureproof the work.

mikeal (approved):

mikeal:
Would love to see one more review approval from someone who works on the gateway.

momack2:
@gmasgras @hsanjuan - could you review please?

BigLep:
We're going this direction, but the work is covered in other proposals including in https://github.com/protocol/web3-dev-team/pull/100 .  Closing as a result.

PR #4, opened 2021-02-16
{=>internal} leverage.

raulk:
Tiny change. The word "leverage" in the context of a product-oriented project definition automatically evokes the question: "what leverage does this project give to our product?"

However, the section really focuses on how executing on this will prepare us to execute better on future endeavours. This is internal team-focused leverage.

### DIFF
--- a/proposals/TEMPLATE.md
+++ b/proposals/TEMPLATE.md
@@ -45,7 +45,7 @@ Explain why you have chosen this rating
 What awesome potential impact/outcomes/results will we see if we nail this project?
 -->
 
-#### Leverage
+#### Internal leverage
 _How much would nailing this project improve our knowledge and ability to execute future projects?_
 
 🎯🎯🎯 = 0-3 emoji rating


proposals/TEMPLATE.md
```diff
@@ -45,7 +45,7 @@ Explain why you have chosen this rating
 What awesome potential impact/outcomes/results will we see if we nail this project?
 -->
 
-#### Leverage
+#### Internal leverage
```
momack2, line 48: Maybe "Dev Velocity"? internal leverage is jargon to me...

raulk: Hmmm... this is about `Acceleration`?

anorth: Agree with @raulk 's interpretation and desire to change this. It's not only about velocity though. Learning and information is an equally or more important aspect. I believe that product-market fit canon emphasises optimising for rate of learning, especially in early stages.

I would actually be happy to change this to "Learning", given that velocity might be captured in Impact.

Or how about "Learning and velocity"?

jacobheun: The counter argument here is to just remove this section entirely in favor of impact. Right now impact is aimed at being solely oriented towards PMF and users, but projects can also have a high impact by getting us more information or enabling us to move more quickly. I don't quite get the need to separate the two, my assumption is that we're trying to be able to quickly clarify what has direct and indirect impact. However, with the removal of quantification, https://github.com/protocol/web3-dev-team/pull/20, this feels like an unnecessary distinction and I recommend we simplify to just having an `Impact` section.

warpfork:
Agree with the original problem statement that it's easy to use the word "leverage" in sentences like "what leverage does this project give to our product?", and that this probably a warping of the intended effect.

Strongly like making space to discuss impact and leverage separately, whatever we call them.  (Perhaps because I often propose things which I suspect are not high impact but are high leverage!)  Suspect if we ditch one of these it'll make a large body of work (anything that's not user-facing) start hiding out of sight of this proposal process, which is not desirable.  (Worth noting that I got to this issue by asking the question "should developers be avoiding putting non-user-facing work into this proposal process?", and the answer was "no!  put reasons for that work in the 'leverage' field".  Fact that I asked that question, on and did so in response to confusion expressed by several other folks in various venues, suggests there's high value to reserving a place for both kinda of evaluation... we just need to increase the clarity of that.)

Not very attracted to "velocity", that term's never really done it for me.  The physics metaphors for programming agility are... okay, so what we really want is "specific impulse enhancements" right?  :trollface: 

...

Maybe an alternative solution path to the problem identified here is having a tiny glossary which hammers home what the words should mean, rather than trying to find a single word which is immune to misinterpretation?

mikeal:
At this point the discussion about this is much longer than the text change, we should either merge this and iterate or close it and propose a larger change in another PR.

PR #6, opened 2021-02-16
#06 POC Projects and Tutorials

zebateira:
Developing POC Projects along with tutorials can provide proof that our stack works and how devs can go ahead and use it for their specific use cases.

This proposal creates a streamlined process for creating POCs and respective tutorials that explain what and how the projects were created and how our stack contributed to it.

An example of this is [Share IPFS](https://share.ipfs.io). Possible tutorial would be "How to create a serverless wetransfer alternative using IPFS".

### DIFF
--- a/proposals/06-poc-projects-with-tutorials.md
+++ b/proposals/06-poc-projects-with-tutorials.md
@@ -1,0 +1,110 @@
+# POC Projects with Tutorials
+
+Authors: zebateira
+
+Initial PR: #6
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+
+It's hard for developers to start using web3 tech and contextualizing IPFS, libp2p and Filecoin in their own needs can go a long way to make them feel at home.
+Building a hello world cli chat app is a good fun exercise, but it's not the end game.
+
+For more complex use cases, it might not be easy to understand if our stack can really pull it off. We need to put it to the test, show everyone how we did it and how they can do it to.
+
+Developing POC Projects along with tutorials can provide just that: proof that our stack works and how devs can go ahead and use it for their specific use cases.
+An example of this is [Share IPFS](https://share.ipfs.io).
+
+Tutorials could build on top of each other by incrementally introducing new parts of our stack. As an example, the part one of a tutorial could start with communication using libp2 and part two of the tutorial could add IPFS integration to store files.
+
+#### Assumptions &amp; hypotheses
+
+- Mature enough tech that can be used to build POC apps
+  - This probably means that libp2p and IPFS are more preferable targets to start with for now as opposed to Filecoin
+- Tutorials in form of articles are a very good entry point for developers to read on new ways to build apps
+
+#### User workflow example
+
+Developers would most likely start with the tutorial/article that details how we can create X app using libp2p and/or IPFS.
+The app should be deployed and publicly accessible for devs to play with: this is the proof that our stack works so it needs to be fairly bug-free and stable.
+
+#### Impact
+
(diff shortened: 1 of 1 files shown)


jessicaschilling:
I like the idea of using share.ipfs.io as an initial effort for this:
- It's a recently updated (and therefore not risky/outdated/buggy) project
- It may not be explicitly FIL-related, but it's _done_, which enables us to concentrate on learning how to write the best sort of PoC tutorial without having to build the PoC code for this initial effort

proposals/06-poc-projects-with-tutorials.md
```diff
@@ -0,0 +1,110 @@
...
+#### What does done look like?
+
+1. First POC Project and Tutorial built and published to show the impact
+1. Detailed guide for building more POC Projects and Tutorials
+    - Would focus on making sure the projects are a good fit for this purpose and guide them on specific nuances of building these apps and tutorials.
+1. UI kit guide or even a design system to reduce visual UI implementation to almost zero.
+1. Process for people to submit ideas, build and publish.
+1. Have a plan for releasing the POCs and tutorials on a frequent schedule.
+1. Streamlined process for analyzing impact with specific metrics
+    - Since we will own the apps, we can easily track the user journey from the tutorial to the app and request feedback from the user.
```
momack2, line 74: Can you explain what you mean by this? If other folks are building their own versions of each app - where do we collect the usage data? (Note that article sessions / code fetches could probably be sufficient here)

proposals/06-poc-projects-with-tutorials.md
```diff
@@ -0,0 +1,110 @@
...
+
+
+####  What does success look like?
+
+POC apps get good traffic and response from the tutorials are mostly positive.
```
momack2, line 79: The ultimate goal we’re shooting for here is successful developers getting great value from the stack. Therefore I’d frame success in “time to build a capable POC”, developer pass through vs bounce, and number of devs flowing through this tutorial (sessions, successful completions, etc).

momack2: I’d make it a non-goal to create such good POCs that they get organic usage/communities and focus instead of them being a great onboarding tool!

momack2 (commented):
Love this proposal and the idea of using share.IPFS.io for it! A few comments about getting more precise about measuring project success

proposals/06-poc-projects-with-tutorials.md
zebateira: > other folks are building their own versions of each app

That is a limitation, but I was referring to the things we _can_ track.

Does this make more sense:

```suggestion
    - Since we will own the apps, we can easily track the user journey from the tutorial to the app and request feedback from the user.
    - We can track tutorial links that go to our deployed app (the app itself will have tracking) and maybe even feedback surveys (if they make sense in the context).
    - We can then encourage developers to fork the app repo from both the tutorial and the app with a fork-me button
    - The number of forks in the repo will be the indicator of the last step of the user completing the tracking flow.
```

proposals/06-poc-projects-with-tutorials.md
zebateira: Makes sense! What do you think about this:

```suggestion
- Time to build a capable POC is shortened
- Positive developer adoption by reducing bounces
- Good completion rate for completing the tutorial

```

mikeal:
Is the next step here to 👍👎 this proposal, or is the next step to assign a specific POC to this proposal and then 👍👎?

PR #7, opened 2021-02-16
#07 Datastore optimization

willscott:


proposals/07-datastore-optimization.md
```diff
@@ -0,0 +1,101 @@
...
+## Purpose &amp; impact 
+#### Background &amp; intent
+
+Chain syncing in Filecoin is painfully slow. This affects the whole product -- system requirements to run even a client node are very high, and gas prices and message limits are limited by the speed we can sync the chain. One huge source of this time suck is simply reading data in and out of our datastore.
+
+We have approached data persistence through using off the shelf key-value stores in our stack. Long term, many aspects of our stack are limited by the lack of a datastore/database designed specifically for our data format, IPLD (see this [overview](https://docs.google.com/document/d/1GxUNUaBfaY4HneWCOfyPcyWG_14t-vnsy3NbQRF4Fdc/edit) and [analysis](https://docs.google.com/document/d/16_8TEslIY__54x2lvMNu5dLL0BSAJ2GxkZ9w3p8sUys/edit)). But in the near term, we have several extremely promising alternative key/value store solutions that can deliver immediate concrete improvements to chain performance.
+
+The aim here is to finalize existing explorations to deliver at least a 50% improvement to chain performance.
+
+#### Assumptions &amp; hypotheses
```
momack2, line 17: These are great. Agree there are risks here in this project’s scope and reliability at scale (would love comments from @magik6k @raulk and @anorth on these risks)

proposals/07-datastore-optimization.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### Leverage
+
+🎯🎯
+
+We are specifically picking this as an easy win on the datastore improvement track. However, it might be something we bring to IPFS eventually, and it will set us up to build more specialized IPLD solutions.
+
+#### Confidence
+
+3 - All user reports that the hardware requirements for running a full lotus node are a major pain point -- if we could drop those requirements, even a little, it would make a big difference
```
momack2, line 44: Totally agree with this confidence in the problem - but as per the really good flags above, we aren’t quite as confident in our proposed solution

proposals/07-datastore-optimization.md
```diff
@@ -0,0 +1,101 @@
...
+We have three significant potential solutions that are well prototyped and close to delivery --
+ * [Using a custom append only key value store](https://github.com/filecoin-project/lotus/pull/5575)
+ * [Split store](https://github.com/filecoin-project/lotus/pull/4992)
+ * [Archival store](https://github.com/ribasushi/ltsh/tree/psql_fullcapture)
+
+The plan of attack is to aggressively test these solutions to determine the best combination of data store, including potentially a segregated choice between hot and cold storage, and deliver a solution that can become the default for Lotus.
+
+* Lotus CLI extended to allow a command line flag selection of datastore
+* Support code for split store, custom key-value, and archival chain store as in-repo options so they are maintained
+* Tests of functionality for all backends to mitigate future breakage / code rot.
```
momack2, line 58: 🙏👍👏💪

proposals/07-datastore-optimization.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### What does done look like?
+
+A released version of lotus with faster chainsync
+
+####  What does success look like?
+
+Success means drop in gas fees as more messages fit in a block
+
+Success means being able to slightly lower system requirements to run a lotus full node
```
momack2, line 68: Would this also include being able to sync the chain faster, or just a sync’d chain taking up less disk space?

proposals/07-datastore-optimization.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+#### Counterpoints &amp; pre-mortem
+
+* Debugging and resolving all issues with proposed solutions could take a while
+* Chain sync speed will likely still be hampered by other factors (CPU load, etc)
+* We don’t have good benchmarks already, which means we have only anecdotal evidence that these solutions will work
+* If we add a custom key value store that could become a significant point of maintenance work
```
momack2, line 76: +1 (@magik6k please opine from stewards end)

momack2 (commented):
Great proposal. Love the call out to assumptions/risks! Let’s get some feedback from lotus stewards

proposals/07-datastore-optimization.md
hannahhoward: faster

proposals/07-datastore-optimization.md
magik6k: I think the point here is that if we write a custom kv store we commit ourselves to maintaining it, which, especially in the beginning will require significant effort

anorth:
Solid proposal, and I gather the splitstore approach is further along now than when this was written.

I would be very wary of the approach of writing/maintaining our own k/v store as the near-term, low-cost solution. It is very easy to underestimate the ongoing pain and cost likely to be associated with that, even after taking into account how easy it is to underestimate! Even if it's "done" and looks like it works, I'd caution against relying on it.

I'm more enthusiastic about that reinvention only when it's considered as a larger project, so that the costs are amortized over a much larger scope of functionality (e.g. as in my IPLD Datastore pitch).

mikeal (approved):

mikeal:
Would love to see one more review approval, but LGTM.

PR #12, opened 2021-02-16
Proposal: F3 - Filecoin storage deals reliability and UX

raulk:

PR #16, opened 2021-02-17
Proposal: Reduce friction and time to first verified deal for new Filecoin storage clients

terichadbourne:
This proposal suggests a path to carry on with the great work started by @andrew & @jnthnvctr's Filecoin storage attempts, addressing the issues flagged with a combination of engineering and documentation. It would also serve as a test ground for establishing a workflow for communicating and collaborating across teams within W3DT as we tackle fundamental engineering issues and documentation in tandem.

### DIFF
--- a/proposals/first-filecoin-storage-deal.md
+++ b/proposals/first-filecoin-storage-deal.md
@@ -1,0 +1,130 @@
+# Proposal: Reduce friction and time to first verified deal for new Filecoin storage clients
+
+Authors: @terichadbourne
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
+A proposal should contain enough detail for others to understand how this project contributes to our team’s mission of product-market fit
+for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on,
+and any other information relevant for prioritizing this project against others.
+It does not need to describe the work in much detail. Most technical design and planning would take place after a proposal is adopted.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). 
+Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+As demonstrated in recent walkthroughs by Andrew Nesbitt and Jonathan Victor, there are a number of critical hurdles blocking an easy path to storing files on Filecoin, including both engineering issues (e.g. Powergate doesn't work), UX issues (lack of description in CLI about what stages exist, showing miners who aren't actually accepting deals, etc.), and documentation issues (incorrect or incomplete instructions for steps in the process that otherwise are functional).
+
+Since our goal is to promote use of the full W3DT stack by developers with storage needs (as opposed to miners), and we aim to include Filecoin in that journey, most paths to adoption are blocked until it's possible to quickly and easily make one's first storage deal. 
+
+#### Assumptions &amp; hypotheses
+- It should be possible to store a file on Filecoin
+- Documentation should accurately reflect the recommended actions
+- Developers who attempt to use our product now and fail are unlikely to trust us in the future, once the experience has improved
+
+#### User workflow example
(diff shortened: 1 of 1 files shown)


rvagg:
I think the framing of this project should still be exploratory, even though we've had some exploratory work done (i.e. step 1 being "done" might not be quite right). The scope is potentially _huge_ since solving it might involve touching many parts of the technology stack - i.e. it doesn't seem to be just a "the documentation isn't good enough" task.

I'm suspecting there's overlap here between the skills and interests of the dev onboarding and dev optimisation (sudo) teams . It might be something we could tackle together in some way, or divide up and treat it as educational for us and expect it to yield further TODOs (project proposals) along the way. Unless PMs (@jnthnvctr?) have confidence already that we know what needs to be solved and we can go ahead and spin off more discrete project proposals?

#12 is related in fixing some of the deal flow pain.

alanshaw:
We have some overlap: https://github.com/protocol/web3-dev-team/pull/28

Maybe this is the super-proposal of mine.

atopal:
@rvagg Yeah, the potential scope is very large, so it might make sense to either use this as the description of the overall desired outcome or scope more tightly. That said, documentation can cover workarounds where the implementation is not sufficient, eg recommending hosted solutions for the time being, and product changes can be on the level of "more comprehensible output of the CLI" or better default behavior.

proposals/16-first-filecoin-storage-deal.md
```diff
@@ -0,0 +1,130 @@
+# Proposal: Reduce friction and time to first verified deal for new Filecoin storage clients
```
ZenGround0, line 1: If I understand this proposal correctly the scope is basic filecoin deals of the type Andrew and Jonathan's walkthroughs attempted to make, not deals related to the ongoing FIL+ (i.e. verified clients/verified deals) effort.  If this is right then the term "verified deal" unfortunately has some extra baggage you probably didn't intend to bring along.  Possible wording changes: "first completed deal", "first finalized deal"

terichadbourne: Thanks for calling that out, @ZenGround0. You're right that it wasn't my intention. I just remember that the process for getting a success message is somewhat long, so the goal is getting to the stage at which the user is confident their data is being stored by a miner.

proposals/16-first-filecoin-storage-deal.md
```diff
@@ -0,0 +1,130 @@
...
+## Project definition
+#### Brief plan of attack
+
+Step 1: _DONE!_ Have sample users attempt to store a file on Filecoin by walking through existing available documention, with no help from fellow Labbers. 
+
+Step 2: PM creates a system for tracking and assinging the critical tasks on this path, using Andrew's video and [JV's slides](https://docs.google.com/presentation/d/1UbO7LKo47KTZcPL8xK41ubFEZwJ7MeFnnAT0_y1yIII/edit?usp=sharing) as a starting point, particularly slide 42 where he outlines which problems can be solved by documentation and which require engineering (and presumably subsequent changes to documentation).
+
+Step 3 (Docs): Docs writer fixes the documentation pain points by adding more details, ensuring recommendations are tailored to storage clients (not miners), etc. Flags additional pain points found along the way. 
+
+Step 3 (Eng): Engineers tackle identified issues with either core protocol implementation for dealmaking or user experience needs in the CLI (progress indicators, etc.). Flags changes made that will require more updates to docs.
```
momack2, line 64: I think Raul/Dirk’s proposal on deal success interface encompasses a large chunk of this eng work (and should likely be done first)

proposals/16-first-filecoin-storage-deal.md
```diff
@@ -0,0 +1,130 @@
...
+
+Since our goal is to promote use of the full W3DT stack by developers with storage needs (as opposed to miners), and we aim to include Filecoin in that journey, most paths to adoption are blocked until it's possible to quickly and easily make one's first storage deal. 
+
+#### Assumptions &amp; hypotheses
+- It should be possible to store a file on Filecoin
+- Documentation should accurately reflect the recommended actions
+- Developers who attempt to use our product now and fail are unlikely to trust us in the future, once the experience has improved
+
+#### User workflow example
+A developer using Filecoin for the first time would visit docs.filecoin.io and successfully follow instructions there to acquire Filecoin to use for the deal, sync or copy the chain (with an understanding of trade-offs in this decision), view available miners with useful information about pricing and availability for storage deals, negotiate a deal, transfer their file, and receive confirmation that the data has been stored successfully.
```
momack2, line 35: Good lifecycle description! Agree we need docs that walk through the whole journey step by step

momack2 (commented):
Agree this is super important, but I think we should make some deal-making protocol changes first to then flow into a great documentation page on the deal making lifecycle

terichadbourne:
My [latest commit](https://github.com/protocol/web3-dev-team/pull/16/commits/089040bf3d9707a11b67a33789494dc5152e00a2) (at the request of @atopal & @raulk) updates the scope of this proposal to include only documentation of current best practices, not engineering of improved processes or CLI UX. (That engineering work would take place within the scope of #12 instead.) It also adds documentation of retrieval best practices rather than focusing exclusively on the storage deal.

PR #17, opened 2021-02-17
Create 17-educate-ipfs-filecoin-intersection.md

terichadbourne:
This proposal aims to battle internal and external confusion about the intersection of IPFS & Filecoin, identify and prioritize engineering needs for user-friendly interop (eg use the CID of a file already on IPFS to refer to it while storing/retrieving on Filecoin), unblock Labbers focused on education from confidently discussing the connection between the projects, and conduct an educational media blitz to get the truth out there!

### DIFF
--- a/proposals/17-educate-ipfs-filecoin-intersection.md
+++ b/proposals/17-educate-ipfs-filecoin-intersection.md
@@ -1,0 +1,195 @@
+# Proposal: Educate internally and externally on the intersection of Filecoin & IPFS
+
+Authors: @terichadbourne
+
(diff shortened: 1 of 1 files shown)


mikeal (approved):

mikeal:
This proposal sufficiently express the goals and scope of this work. My only concern is that we don’t have this story entirely worked out and we’re still considering projects and protocol improvements that change how this works. We may want to wait for those projects to shake out before we take this on.

BigLep:
Still waiting as engineering implementation work progresses and settles.

PR #18, opened 2021-02-17
Symbiosis protocol pitch

Gozala:
cc @hugomrdias

### DIFF
--- a/proposals/Symbiosis Protocol.md
+++ b/proposals/Symbiosis Protocol.md
@@ -1,0 +1,387 @@
+# Symbiosis protocol
+
+<!--
+This template is intended to be used by those who would like to pitch a new project for one of the Web3 Dev project teams to take on. It should contain sufficient detail that others can understand how this project contributes to our team’s mission of  product-market fit for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on, and any other information relevant for prioritizing this project against others.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+
+### Purpose &amp; impact 
+##### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+There is an inherent asymmetry in node capabilities imposed by the [constraints of the operating environment](##constraints-of-the-operating-environment). Attempts to overcome these limitations often lead to [ad-hoc stop-gap solutions that introduce centralization](#ad-hoc-stop-gap-solutions) into the system. Teams building products on IPFS are required to fill capability gaps with [(unintentionally) incompatible infrastructure](#Unintentionally-Incompatible-infrastructure) of their own.
+
+[IPFS HTTP API provides inadequate solution](#HTTP-API-provides-inadequate-solution) for constrained environments as it is designed to give a single client an **absolute control** of the host node as opposed to coordinating access of multiple tenants.
+
+Here we propose design which embraces asymmetry among node capabilities. Instead of traditional client-server model, symbiotic network is formed through a protocol that enables IPFS nodes with limited capabilities and/or lifespan _(from now on referred as **symbionts**)_ to partner with IPFS nodes that posses greater capabilities and/or lifespan _(from now on referred as **hosts**)_.
+
+<!-- In this design symbionts are a hybrid between IPFS Client and an IPFS node with limited capabilities _(Such as IPFS node embedded in a web site)_. In solitude symbionts operate in limited capacity. They can seamlessely upgrade thier capabilities through a partner host. -->
+
+Through the [symbiosis protocol](#Symbiosys-protocol) a **host** can announce [services](#Services) that **symbionts** can _request_ to overcome own limitations. 
+
+Protocol defines `m:n` relationship implying that a symbiont can have multiple partner hosts and a host could partner with multiple symbionts.
+
+
+##### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+- Neither embedded IPFS node nor IPFS client provide an adequate solution for building products in constrained environments.
+- OS level node sharing with existing building blocks is impractial.
+- Lack of generalized delegation protocol between nodes leads ad-hock stop-gap solutions.
+- Teams are unable to plug domain specific extensions without having to fork our stack.
+- Proposed solution avoids accidental centralization (0 cost in host symbiont switching)
+- Provides a seamless capabilities upgrade and downgrade experience.
+
+##### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+#### Example: Symbiont leveraging multiple hosts
+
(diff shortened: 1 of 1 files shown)


proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,387 @@
...
+
+[IPFS HTTP API provides inadequate solution](#HTTP-API-provides-inadequate-solution) for constrained environments as it is designed to give a single client an **absolute control** of the host node as opposed to coordinating access of multiple tenants.
+
+Here we propose design which embraces asymmetry among node capabilities. Instead of traditional client-server model, symbiotic network is formed through a protocol that enables IPFS nodes with limited capabilities and/or lifespan _(from now on referred as **symbionts**)_ to partner with IPFS nodes that posses greater capabilities and/or lifespan _(from now on referred as **hosts**)_.
+
+<!-- In this design symbionts are a hybrid between IPFS Client and an IPFS node with limited capabilities _(Such as IPFS node embedded in a web site)_. In solitude symbionts operate in limited capacity. They can seamlessely upgrade thier capabilities through a partner host. -->
```
alanshaw, line 21: Was this intended to be commented out?

proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,387 @@
...
+- Proposed solution avoids accidental centralization (0 cost in host symbiont switching)
+- Provides a seamless capabilities upgrade and downgrade experience.
+
+##### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+#### Example: Symbiont leveraging multiple hosts
+
+A symbiont on a phone may be paired with a host on a laptop to leverage improved capabilities when the two are on a same wifi network. At the same time a symbiont on the phone can be paired with a remote host over the internet to allowing all user devices to sync on the go.
```
alanshaw, line 45:
```suggestion
A symbiont on a phone may be paired with a host on a laptop to leverage improved capabilities when the two are on a same wifi network. At the same time a symbiont on the phone can be paired with a remote host over the internet to allow all user devices to sync on the go.
```

proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,387 @@
...
+3. Implement permission handling flow on the host.
+4. Make symbiont leverage announced service through the protocol.
+5. Collaborate with teams to validate design with differnt use cases.
+6. Take learnings and implement zeroconfig service discovery.
+
+##### What does done look like?
+_What specific deliverables should completed to consider this project done?_
+
+- Have at least one collab adopt symbiosis protocol.
+- IPFS Desktop providing services that enhances in-browser node capabilities.
```
alanshaw, line 130: Is this possible?

proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,387 @@
+# Symbiosis protocol
```
Gozala, line 1:
```suggestion
# Symbiosis protocol

Authors:
  - [@gozala](https://github.com/gozala)
  - [@hugomrdias](https://github.com/hugomrdias)

Initial PR: https://github.com/protocol/web3-dev-team/pull/18
```

proposals/Symbiosis Protocol.md
Gozala: We decided to get rid of some sections, but instead of deleting commented them in case received feedback would suggest we need this content.

proposals/Symbiosis Protocol.md
Gozala: Yes, if implemented. But question suggest maybe we need to elaborate more on this ?

proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,393 @@
...
+  > // **TODO:** Check how does [IPFS Desktop], and [Space Daemon][] get along with one another
+- Using [IPFS HTTP API][] for sharing would still inherit all the issues.
+
+## Symbiosys protocol
+
+> Protocol draws it's inspired from [DNS Service Discovery (DNS-SD)][DNS-SD] a [zero-configuration networking][] technique.
+
+Wire protocol that is agnostic of transport and representation (message encoding) _(This makes it a good fit for cross thread communication and a cross network interface)_.
+
+At the high level protocol allows IPFS nodes to announce services they can provide to other nodes on the network, which can be requested by connected nodes to overcome their limitations.
```
lidel, line 331: I feel the "Limited Relay" concept from  @aarshkshah1992's   Project Flare (#23) could fit here.
Worth mentioning it as potential use case.

mikeal (approved):

mikeal:
The proposal looks good but I question the confidence. Once we start implementing this I’d expect to surface many unknown unknowns.

proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,393 @@
...
+   5. Host is free to revoke permission
+      > Revokation reason can be arbitrary. E.g. 
+      > - Payment for service was not received.
+      > - In IPFS Desktop user revokes permission to specific web page). 
+
+## Services
+
+Protocol does not prescribe specific set of services with an assumbtion that they could be developed as needed. That said below are couple of examples:
+
+- Enhanced dialability (tunneled through the host)
```
Stebalien, line 357: Can also leverage the host to offload encryption (when trusted).

Stebalien (commented):
This is going to need a _lot_ of careful design work, but I think this is a great idea. Really, I think this is critical for PMF because:

1. It allows users to _extend_ IPFS without having to fork it.
2. It allows users to _use_ IPFS within their application without having to bundle it.

proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,393 @@
...
+
+
+##### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+- Neither embedded IPFS node nor IPFS client provide an adequate solution for building products in constrained environments.
+- OS level node sharing with existing building blocks is impractial.
+- Lack of generalized delegation protocol between nodes leads ad-hock stop-gap solutions.
+- Teams are unable to plug domain specific extensions without having to fork our stack.
```
vasco-santos, line 41: What type of domain specific extensions we are describing here? Dapp developers should create libp2p protocols

proposals/Symbiosis Protocol.md
```diff
@@ -0,0 +1,393 @@
...
+For a team of 3-5 people with the appropriate skills:
+- Small, 1-2 weeks
+- Medium, 3-5 weeks
+- Large, 6-10 weeks
+- XLarge, >10 weeks
+Describe any choices and uncertainty in this scope estimate. (E.g. Uncertainty in the scope until design work is complete, low uncertainty in execution thereafter.)
+-->
+
+- Large
+- Uncertainty on discovery implementation until several services are implemented.
```
vasco-santos, line 185: We will probably need to rely on manual configuration first, even though we can add some defaults like trying default IPFS Desktop address port pair

vasco-santos (commented):
This proposal is really interesting and I think it might make sense to a lot of use cases, and also to solve some of the pain points of our stack.

This will be a massive project full of uncertainty. Perhaps, the plan of action should be decoupled in smaller end to end projects for experimentation, like the delegate protocol on libp2p.

proposals/Symbiosis Protocol.md
Gozala: There are bunch cases where domain specific replication strategy would be far more effective than bitswap is. In fact IPLD selectors is one example where that could be expressed a service. Another one I was thinking is textile threads, where communicating the a head you've seen is enough for other participants to know all the blocks they need to send you.

I think libp2p protocols are indeed relevant here. I think of this as a bit more higher level equivalent and designed with asymmetry in mind.

BigLep:
Assigning to grants since it won't get worked on in the short term with current program structure.  @mikeal is going to evaluate grant possibility.

PR #19, opened 2021-02-17
Reliable Mutability Primitive pitch

Gozala:
cc @hugomrdias @aschmahmann @lidel

### DIFF
--- a/proposals/Reliable Mutability Primitive.md
+++ b/proposals/Reliable Mutability Primitive.md
@@ -1,0 +1,162 @@
+<!--
+This template is intended to be used by those who would like to pitch a new project for one of the Web3 Dev project teams to take on. It should contain sufficient detail that others can understand how this project contributes to our team’s mission of  product-market fit for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on, and any other information relevant for prioritizing this project against others.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+# Reliable Mutability Primitive
+
+### Purpose &amp; impact 
+##### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+
+Today ecosystem has to work around the lack of functional mutability primitives. [IPNS][] does not provide reliable solution forcing community to adopt different custom solutions.
+
+- [Pinata][] gets asked about how to chaneg content for CID onec every few weeks _(Most of the time they find that is not what the users actually need)_.
+- [fission.codes](https://fission.codes/) found IPNS persistance unreliable and had to build custom solution that updates [DNSLink][] records (**TODO**: Get a quote / link from Boris)
+- **TODO**: Survey textile, last time we talked they found IPNS to be unreliable for propagating thread updates.
+- ...
+- Protocol Labs uses dns simple API to [publish site updates](https://github.com/filecoin-project/specs/blob/71f37208a1f4f56b33ea307d7cbdb4b06996b115/.github/workflows/main.yml#L40). 
+
+
+This gap in the stack makes building software which runs on end user computers and meets modern expectations more difficult than traditional hosted solutions. This creates a hardle for adoption of the technology.
+
+
+
+##### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+- Applications want to provide a canonical address for a current state.
+- Application authors will prefer built-in mutability primitive over custom solutions. 
+- Developed solution would provide simpler mechanism for publishing updates (than DNSLink updates)
+- Developed solution would work provide reliable mechanism for producing updates on not always connected end user computers (think browsers, phones)
+- Solution should not require shipping secret keys or tokens with devices.
+
+##### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+Team builds an application enabling user to write and share notes with each other. They chose to use IPFS as that enabled them to focus on the applications writing exprience free from having to write and maintain backend infrastructure or ensuring data safety. Application stores encrypted notes in the embedded IPFS node and publishes CID to the latest shared notes DAG root via IPNS API. This enables collaborating users to fetch updated notes via embedde IPFS. Application also leverage [Pinning Services API][], enabling their users to choose from growing number of service providers or to bring their own endpoint for better availability.
+
+Choosing IPFS not only reduced an engineering effort, but also made their product fully off-grid capable (thanks to IPNS update propagation over mDNS) with no additional effort on their end. Team often likes to points out that their product has an edge over the competition, not only all user e2e encrypted, but their users can also choose where that data is persisted.
+
+##### Impact
+_How directly important is the outcome to our top-level mission?_
+
+
+🔥🔥🔥 This would enable building competitive software that does not introduce custom server components (Standard [Pinning Services API][] provides data persistance and standard mutability primitive provides updating mechanism).
+
+It would also give an edge to the whole ecosystem, because [there is no step three](https://www.youtube.com/watch?v=YHzM4avGrKI) (as in no servers to run or maintain).
+
+
+<!--
+Explain why you have chosen this rating
+What awesome potential impact/outcomes/results will we see if we nail this project?
+-->
+
+##### Leverage
+_How much would nailing this project improve our knowledge and ability to execute future projects?_
+
+
+<!-- Explain the opportunity or leverage point for our subsequent velocity/impact (e.g. by speeding up development, enabling more contributors, etc)
+-->
+
+🎯🎯 - This would enable projects that need to have a canonical identifier for evolving dataset. It would also provide a greater interop across products built around IPFS by removing need for custom incompatibile solutions that teams need to implemente today.
+
+
+##### Confidence
+_How sure are we that this impact would be realized? Label from [this scale](https://medium.com/@nimay/inside-product-introduction-to-feature-priority-using-ice-impact-confidence-ease-and-gist-5180434e5b15)_.
+
+<!--Explain why this rating-->
+
+3 - As high confidence as we can have without doing actual user studies. We know teams that have tried IPNS but found it unrelaible, ultimately rolling out an alternative solution. We also know teams that have evaluated IPFS but chose alternative due to lack of reliable mutability story.
+
+
+
+### Project definition
+##### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+- Implement third party IPNS republishing (TODO: Find a relevant issue link)
+- Collaborate with Pinning Services to make this part of the API.
+- Fix IPNS publishing in browsers.
+- Enable pubsub by default + IPNS over pubsub.
+- Implement & deploy strategic republishing system on PL hosted nodes that will republish popular content and drop stalled one. So that IPNS works even thought pinning services.
+- (Nice to have) Implement and rollout delegated publishing to enable updating from multiple devices.
+- (Nice to have) Implement and rollout versioning system so that nodes can see full update history.
+
+##### What does done look like?
+_What specific deliverables should completed to consider this project done?_
+
+Publishing IPNS record takes just a few miliseconds. Shutting laptop and loading `https://ipfs.io/ipns/{key}` week later on other device loads the content published as fast as loading corresponding `https://ipfs.io/ipfs/{cid}`.
+
+
+#####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+People and teams are storing IPNS keys in [DNSLink][] records as opposed to CIDs, because that provides a more effective and simpler way to publish updates. We see new projects are leveraging IPNS over working around it via things like [DNSLink][].
+
+
+##### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+- IPNS has bad reputation, changing it may prove to be a challenge.
+- Teams may find that general primitive is not adequate for custom needs.
+- Required key management (e.g. key recovery, key compromise) still may prove it impractical.
+- Higher level alternatives like textile threads may prove to be better layer of abstraction.
+
+##### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+- Integrating textile threads as core component of the ecosystem.
+- Standardizing DNS record updates mechanism and intergating into IPFS.
+- Work with Pinning Services to implement stardized DNSLink updating mechanism.
+
+##### Dependencies/prerequisites
+<!--List any other projects that are dependencies/prerequisites for this project that is being pitched.-->
+
+- Functioning IPNS in browsers
+- Enabling pubsub in go-ipfs
+
+##### Future opportunities
+<!--What future projects/opportunities could this project enable?-->
+
+- [Petname][] system for memorable but non canonical names (kind of how our phones allow us to not remember numbers).
+- Subscribtion based deny lists (node can subscribe to any endpoint)
+- RSS like systems (In RSS readers memorable namse were not a concern)
+- DAGs revisions carrying IPNS key so that last latest revision or any past revision could be discovered.
+
+### Required resources
+
+
+##### Effort estimate
+<!--T-shirt size rating of the size of the project. If the project might require external collaborators/teams, please note in the roles/skills section below). 
+For a team of 3-5 people with the appropriate skills:
+- Small, 1-2 weeks
+- Medium, 3-5 weeks
+- Large, 6-10 weeks
+- XLarge, >10 weeks
+Describe any choices and uncertainty in this scope estimate. (E.g. Uncertainty in the scope until design work is complete, low uncertainty in execution thereafter.)
+-->
+
+Large. Requires lot of coordinated changes across implementations and collabs. Uncertainty in the scope until concrete actionable plan is in place. Uncertenty in collab buy-in. 
+
+##### Roles / skills needed
+<!--Describe the knowledge/skill-sets and team that are needed for this project (e.g. PM, docs, protocol or library expertise, design expertise, etc.). If this project could be externalized to the community or a team outside PL's direct employment, please note that here.-->
+
+- go-ipfs development
+- js-ipfs development
+- cross org coordinator (to get pinning services onboard)
+
+[Pinata]:https://pinata.cloud/
+[DNSLink]:https://docs.ipfs.io/concepts/dnslink/
+[IPNS]:https://docs.ipfs.io/concepts/ipns/
+[Pinning Services API]:https://ipfs.github.io/pinning-services-api-spec/
+[Petname]:https://en.wikipedia.org/wiki/Petname


proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,162 @@
...
+##### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+
+Today ecosystem has to work around the lack of functional mutability primitives. [IPNS][] does not provide reliable solution forcing community to adopt different custom solutions.
+
+- [Pinata][] gets asked about how to chaneg content for CID onec every few weeks _(Most of the time they find that is not what the users actually need)_.
```
rvagg, line 17: minor typos - `change`, `once`

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,162 @@
...
+- [fission.codes](https://fission.codes/) found IPNS persistance unreliable and had to build custom solution that updates [DNSLink][] records (**TODO**: Get a quote / link from Boris)
+- **TODO**: Survey textile, last time we talked they found IPNS to be unreliable for propagating thread updates.
+- ...
+- Protocol Labs uses dns simple API to [publish site updates](https://github.com/filecoin-project/specs/blob/71f37208a1f4f56b33ea307d7cbdb4b06996b115/.github/workflows/main.yml#L40). 
+
+
+This gap in the stack makes building software which runs on end user computers and meets modern expectations more difficult than traditional hosted solutions. This creates a hardle for adoption of the technology.
```
alanshaw, line 24:
```suggestion
This gap in the stack makes building software which runs on end user computers and meets modern expectations more difficult than traditional hosted solutions. This creates a hurdle for adoption of the technology.
```

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+##### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+- Applications want to provide a canonical address for a current state.
+- Application authors will prefer built-in mutability primitive over custom solutions. 
+- Developed solution would provide simpler mechanism for publishing updates (than DNSLink updates)
+- Developed solution would work provide reliable mechanism for producing updates on not always connected end user computers (think browsers, phones)
```
alanshaw, line 35:
```suggestion
- Developed solution would provide a reliable mechanism for producing updates on not always connected end user computers (think browsers, phones)
```

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,162 @@
+<!--
```
Gozala, line 1:
```suggestion
# Reliable Mutability Primitive

Authors: [@gozala](https://github.com/gozala)

Initial PR: https://github.com/protocol/web3-dev-team/pull/19/
<!--
```

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,162 @@
...
+This template is intended to be used by those who would like to pitch a new project for one of the Web3 Dev project teams to take on. It should contain sufficient detail that others can understand how this project contributes to our team’s mission of  product-market fit for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on, and any other information relevant for prioritizing this project against others.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+# Reliable Mutability Primitive
```
Gozala, line 5:
```suggestion
```

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+3 - As high confidence as we can have without doing actual user studies. We know teams that have tried IPNS but found it unrelaible, ultimately rolling out an alternative solution. We also know teams that have evaluated IPFS but chose alternative due to lack of reliable mutability story.
+
+
+
+### Project definition
+##### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+- Implement third party IPNS republishing (TODO: Find a relevant issue link)
```
aschmahmann, line 88: This is a closely related issue https://github.com/ipfs/go-ipfs/issues/4435

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+- Collaborate with Pinning Services to make this part of the API.
+- Fix IPNS publishing in browsers.
+- Enable pubsub by default + IPNS over pubsub.
```
aschmahmann, line 91: Note: Textile threads utilize pubsub which means that any issues with enabling pubsub by default are things that would be experienced by Textile threads users. They also utilize IPNS over PubSub for buckets.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+- Implement & deploy strategic republishing system on PL hosted nodes that will republish popular content and drop stalled one. So that IPNS works even thought pinning services.
+- (Nice to have) Implement and rollout delegated publishing to enable updating from multiple devices.
```
aschmahmann, line 93: This is sort of doable already "for free" if we have some way to pin/follow IPNS updates. I can just share the private key with my multiple devices via `ipfs key export/import`.

There are alternative approaches too, but figured I'd point this one out.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+Publishing IPNS record takes just a few miliseconds. Shutting laptop and loading `https://ipfs.io/ipns/{key}` week later on other device loads the content published as fast as loading corresponding `https://ipfs.io/ipfs/{cid}`.
+
+
+#####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+People and teams are storing IPNS keys in [DNSLink][] records as opposed to CIDs, because that provides a more effective and simpler way to publish updates. We see new projects are leveraging IPNS over working around it via things like [DNSLink][].
```
aschmahmann, line 108: At the very least because updating DNS is such a pain we see people using DNSLink and pointing it at IPNS records instead of IPFS because its so much easier to work with IPNS updates than DNS updates. Similarly, we may see people storing IPNS addresses in ENS to save on not wanting to publish many blockchain transactions.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+- Integrating textile threads as core component of the ecosystem.
+- Standardizing DNS record updates mechanism and intergating into IPFS.
+- Work with Pinning Services to implement stardized DNSLink updating mechanism.
+
+##### Dependencies/prerequisites
+<!--List any other projects that are dependencies/prerequisites for this project that is being pitched.-->
+
+- Functioning IPNS in browsers
```
aschmahmann, line 129: Note: likely includes an up-to-date IPNS over PubSub implementation

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+### Project definition
+##### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+- Implement third party IPNS republishing (TODO: Find a relevant issue link)
+- Collaborate with Pinning Services to make this part of the API.
+- Fix IPNS publishing in browsers.
+- Enable pubsub by default + IPNS over pubsub.
+- Implement & deploy strategic republishing system on PL hosted nodes that will republish popular content and drop stalled one. So that IPNS works even thought pinning services.
```
aschmahmann, line 92: Interesting, I'm not sure this is really required as we can give IPNS the same guarantees as IPFS. However, since the amount of data is smaller this idea of a public IPNS pinning service seems pretty cool.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+- Enabling pubsub in go-ipfs
+
+##### Future opportunities
+<!--What future projects/opportunities could this project enable?-->
+
+- [Petname][] system for memorable but non canonical names (kind of how our phones allow us to not remember numbers).
+- Subscribtion based deny lists (node can subscribe to any endpoint)
+- RSS like systems (In RSS readers memorable namse were not a concern)
+- DAGs revisions carrying IPNS key so that last latest revision or any past revision could be discovered.
+
```
aschmahmann, line 139: Also, more advanced mutability systems could be built (e.g. where instead of a single public key, we can have M of N signers required on records). But it all starts from having the UX and process around pinning mutable names and having a performant and efficient mutability system (e.g. Persistent PubSub, like what IPNS over PubSub uses)

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+<!--T-shirt size rating of the size of the project. If the project might require external collaborators/teams, please note in the roles/skills section below). 
+For a team of 3-5 people with the appropriate skills:
+- Small, 1-2 weeks
+- Medium, 3-5 weeks
+- Large, 6-10 weeks
+- XLarge, >10 weeks
+Describe any choices and uncertainty in this scope estimate. (E.g. Uncertainty in the scope until design work is complete, low uncertainty in execution thereafter.)
+-->
+
+Large. Requires lot of coordinated changes across implementations and collabs. Uncertainty in the scope until concrete actionable plan is in place. Uncertenty in collab buy-in.
```
aschmahmann, line 153: True, although just getting things stable + performant in go-ipfs and web browsers would likely go along way while collabs conversations are underway.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+
+Today ecosystem has to work around the lack of functional mutability primitives. [IPNS][] does not provide reliable solution forcing community to adopt different custom solutions.
+
+- [Pinata][] gets asked about how to change content for CID once every few weeks _(Most of the time they find that is not what the users actually need)_.
+- [fission.codes](https://fission.codes/) found IPNS persistance unreliable and had to build custom solution that updates [DNSLink][] records (**TODO**: Get a quote / link from Boris)
+- **TODO**: Survey textile, last time we talked they found IPNS to be unreliable for propagating thread updates.
```
Gozala, line 23: [@carsonfarmer: This is still correct, IPNS is too unreliable for anything with essentially sub-minute temporal resolution. So for data updates, it is not useful. However, we are finding it quite useful for buckets, and have used it to great effect recently.](https://hackmd.io/@gozala/ipns-pitch)

proposals/Reliable Mutability Primitive.md
Gozala, line 88: @carsonfarmer: [Yeh, this one shouldn't be ignored... maybe the existing solution is just not going to cut it in the end, and a fresh start is required. How likely is it that we could see "two" official methods, and have one of them win out?](https://hackmd.io/@gozala/ipns-pitch)

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+
+- IPNS has bad reputation, changing it may prove to be a challenge.
+- Teams may find that general primitive is not adequate for custom needs.
+- Required key management (e.g. key recovery, key compromise) still may prove it impractical.
+- Higher level alternatives like textile threads may prove to be better layer of abstraction.
+
+##### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+- Integrating textile threads as core component of the ecosystem.
```
Gozala, line 122: @carsonfarmer: [Obviously I like this idea. But it would end up being something threads-like I think. A benefit here is that threads supports multi-writer updates, which I think its a feature lacking in IPNS right now. The downside is that threads "as is" won't scale. Additionally, the encryption layer in threads is too heavy handed right now for a general solution like this. Having said all that, these things are all fixable implementation details.](https://hackmd.io/@gozala/ipns-pitch)

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+- Standardizing DNS record updates mechanism and intergating into IPFS.
+- Work with Pinning Services to implement stardized DNSLink updating mechanism.
```
Gozala, line 124: @carsonfarmer: [This would be an interesting shorter term solution](https://hackmd.io/@gozala/ipns-pitch)

proposals/Reliable Mutability Primitive.md
Gozala: @aschmahmann do you mean something like 

```suggestion
- Functioning IPNS in browsers (over up-to-date PubSub implementation)
```

Or a separate item ? I

aschmahmann: Well IPNS over PubSub has a few tweaks over just PubSub itself to give some reliability. It's specced [here](https://github.com/ipfs/specs/blob/a370a6aa2345b251a4c04d37b8c50d3aa8f2701f/naming/pubsub.md).

@hugomrdias would probably know more about other work required to get IPNS functioning in a browser I just know that without the IPNS over PubSub changes in the spec that it won't be fast enough.

proposals/Reliable Mutability Primitive.md
carsonfarmer: I should also mention something else to provide further context: We're currently managing 1000s (yes, that's right) of IPNS updates for buckets. This currently takes an extremely long time to republish. We are actively conducting tests to determine exactly how long this takes, and how often it fails to complete. In some cases, it is taking hours per day.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+### Project definition
+##### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+- Implement third party IPNS republishing (TODO: Find a relevant issue link)
+- Collaborate with Pinning Services to make this part of the API.
+- Fix IPNS publishing in browsers.
```
BigLep, line 90: (New person comment): is this essential for unblock the core use-case of enabling customers to do ipns updates instead of DNS updates?

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+People and teams are storing IPNS keys in [DNSLink][] records as opposed to CIDs, because that provides a more effective and simpler way to publish updates. We see new projects are leveraging IPNS over working around it via things like [DNSLink][].
+
+
+##### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+- IPNS has bad reputation, changing it may prove to be a challenge.
```
BigLep, line 114: How do we get more data on this?  Is this a question to PM's?

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+- Teams may find that general primitive is not adequate for custom needs.
```
BigLep, line 115: Have we been presented any usecases where ipns (once it has a reliable mutability primitive) isn't sufficient?

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+- Required key management (e.g. key recovery, key compromise) still may prove it impractical.
+- Higher level alternatives like textile threads may prove to be better layer of abstraction.
+
+##### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+- Integrating textile threads as core component of the ecosystem.
+- Standardizing DNS record updates mechanism and intergating into IPFS.
```
BigLep, line 123: Is this even possible?  Doesn't it depend on each DNS' API for mutating actions?

proposals/Reliable Mutability Primitive.md
BigLep: (new person coming up to speed): what would this look like?  I think I'll be able to grok it with a tiny sketch or a mock sample code example.

proposals/Reliable Mutability Primitive.md
BigLep: Sorry for the basic question, but I'm still confused by what is meant by "functioning IPNS in browsers".  What isn't working browsers today?

proposals/Reliable Mutability Primitive.md
BigLep: @Gozala : can we put some high-level estimates (in terms of dev weeks) around each item in the "Brief plan of attack" above?  I think that will help give a bit more clarity on the size of this project.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
...
+-->
+
+Large. Requires lot of coordinated changes across implementations and collabs. Uncertainty in the scope until concrete actionable plan is in place. Uncertenty in collab buy-in. 
+
+##### Roles / skills needed
+<!--Describe the knowledge/skill-sets and team that are needed for this project (e.g. PM, docs, protocol or library expertise, design expertise, etc.). If this project could be externalized to the community or a team outside PL's direct employment, please note that here.-->
+
+- go-ipfs development
+- js-ipfs development
+- cross org coordinator (to get pinning services onboard)
```
BigLep, line 160: Sorry to be slow, but why is it required to get pinning services onboard?  I'm not saying we should skip doing this, I just want to make sure I understand the inter-dependency.  For example, in my head right now I understand that Pinning Services are critical for persistence but I'm less clear on their role in mutability.

BigLep (commented):
Thanks a lot for putting this together - good stuff.  Apologies if the questions below are all basic.  Feel free to point me anywhere I should be reading to self-service answer.

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,166 @@
+# Reliable Mutability Primitive
+
+Authors: [@gozala](https://github.com/gozala)
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/19/
```
BigLep, line 5: Maybe add a note that this is part of the larger story in https://github.com/protocol/web3-dev-team/pull/42 ?

Gozala: I am not sure this would be accurate. I think there is a merit to reliable mutability primitive whether we decide to do human readable names or not.

proposals/Reliable Mutability Primitive.md
Gozala, line 88:
```suggestion
- Implement third party IPNS republishing (see [ipfs/go-ipfs#4435](https://github.com/ipfs/go-ipfs/issues/4435))
```

proposals/Reliable Mutability Primitive.md
Gozala: Yes if those updates were to made from browser. It is also however useful on it's own merit as it enables apps to pull updated content even if such content has no human readable name.

proposals/Reliable Mutability Primitive.md
Gozala: Yeah I meant getting it to a state where we can comfortably tell users this is how you go about it. I do have some reservations towards key sharing, but that's probably discussion worth having when we get here.

proposals/Reliable Mutability Primitive.md
Gozala, line 108:
```suggestion
- People and teams are storing IPNS keys in [DNSLink][] records as opposed to CIDs, because that provides a more effective and simpler way to publish updates.
- We see people storing IPNS addresses in ENS to save on blockchain transactions costs.
- We see new projects leveraging IPNS (when human readable name is not a concern) instead of working around it wit [DNSLink][].
```

proposals/Reliable Mutability Primitive.md
Gozala: I think it's well know fact, evidence been that teams have tried and ultimately end up going with DNSLink solution. I'm pretty sure PL did that too, but someone with longer history here might know better. 

We could also get specific collabs to comment here to either validate the claim or dispute it. I have invited folks from textile & fission to validate. We could reach out to more teams.

@olizilla do you by a chance know if we used to have IPNS in DNSLink past or am I making wrong assumptions here.

proposals/Reliable Mutability Primitive.md
Gozala: - IPNS lacks human readable/memorable naming which is a concern in most human facing interactions. 
- I have in the past voiced concerns about the lack of revision history associated with it (although one could built such a system on top).
- There also had been discussions about ability to have multiple publishers to the same name.
- And then there are also questions about compromised or lost key recovery


I suspect that despite all this usable IPNS will address enough use cases to be worth it. That said we might find that a higher level primitive that address more use cases out of the box will provide a better market fit.

proposals/Reliable Mutability Primitive.md
Gozala: Hypothetically speaking we can start a standardization effort form a committee  and get enough big players onboard to get it done. I imagine there might be some value for DNS providers to make their products more accessible.

proposals/Reliable Mutability Primitive.md
Gozala: We have recently created [pinning service API](https://ipfs.github.io/pinning-services-api-spec/#section/Schemas) that allows you to add pinning services like pinata to your IPFS node and then remote pin things. Kind of like adding git remote and pushing changes to it.

Idea here is to extend that API in a way that you could update DNSLink provided by the pinning service to the new CID through extended API. Kind of like when you push to remote it updates corresponding branch to new commit. In our case branch names would be DNSLinks, that you could update to point to new CID.

@BigLep does this answer your question ?

proposals/Reliable Mutability Primitive.md
Gozala: I should probably capture this in the document. Currently pubsub is disabled in go-ipfs so all the IPNS records are published and resolved from DHT. Nodes in browsers do not have necessary capabilities to query/publish DHT. Instead they do publish and resolve IPNS records over pubsub. Result is fragmented network where go and js nodes see each others records.

Furthermore because because there is not enough nodes on the network with pubsub enabled published records from browser don't always make to nodes subscribed to them.

P.S.: Some of this may not be completely accurate, @hugomrdias @aschmahmann could you please confirm I'm not misrepresenting current state.

proposals/Reliable Mutability Primitive.md
Gozala: I'm afraid I'm not well qualified to do this. I've mostly put this proposal together from prior discussions I had with @aschmahmann and @hugomrdias (still in outsider capacity) in an attempt to rally folks who have much better understanding of internal details.

Help trying to scope it out further, would be greatly appreciated.

P.S: Me & @hugomrdias plan no going over this proposal tomorrow to flush it out further.

proposals/Reliable Mutability Primitive.md
Gozala: That is a valid point and calling out pinning services specifically might be projecting too far of how things would work. 

This was partially influenced by the fact that IPNS (at least in its current form) requires records to be republished otherwise they'll become unsolvable. Implicit  assumption here, is that, unless there are motivated nodes to do this, system will end up impractical. Pinning services just seem well positions to do this given that they are motivated nodes keep content that IPNS will point to available (insert git branch analogy here).

That said we could and probably should consider different strategies to meet reliability goals.

Gozala: In other words I think we still need someone to help coordinate efforts beyond just leading implementation efforts, to ensure that we do not end up with a system that is directly tied to PL infrastructure. I am assuming this involves getting other teams onboard.

proposals/Reliable Mutability Primitive.md
lidel: @BigLep good pulse check in this issue (2017–today): https://github.com/ipfs/go-ipfs/issues/3860

proposals/Reliable Mutability Primitive.md
BigLep: @Gozala : thanks - this helps.  The only open area for me then is where is this DNSLink stored?  Is this a DNSLink in a DNS record on a (sub)domain that the Pinning Service owns?  

Basically would the chain be
``My domain (e.g., mywebsite.com) has DNS record that points to a domain the Pinning service controls (e.g., mywebsite.somepinningservice.com) which has the DNSLink to the latest CID?``

lidel: @BigLep Depends if you (A) publish to your own domain, or  (B) use subdomain provided by pinning service for free. 

To give you a real life example, let's say  we want to host https://cid.ipfs.io on some pinning service (pin + take care of updating dnslink + TLS).

(B) is easy. Pinning service simply give you a name on some domain they own. Fleek does that already, they give  https://cid-utils-website.on.fleek.co/  for free + they set up DNSLink for those names (because they own the top domain).

(A) means we want to delegate updates of  DNS TXT record on  `_dnslink.cid.ipfs.io` to some pinning service. 
This can be done either by:
- giving DNS provider credentials to pinning service + expecting them to implement support for proprietary APIs of each dns provider (not realistic IMO, but mentioning here for completeness)
-  using DNS protocol itself for delegation: setting NS records that delegate management of `_dnslink.cid.ipfs.io` to a DNS server controlled by pinning service. That way they have full control over DNSLink without having any control over other records (because we delegated only `_dnslink` subdomain.

proposals/Reliable Mutability Primitive.md
lidel: +1 on at least thinking about pinning services in the context of IPNS:

- how pinning of /ipns/{libp2p-key-cid} would work?
- how they can help with avoiding IPNS rot 

etc

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,168 @@
...
+- Fix IPNS publishing in browsers.
+- Enable pubsub by default + IPNS over pubsub.
+- Implement & deploy strategic republishing system on PL hosted nodes that will republish popular content and drop stalled one. So that IPNS works even thought pinning services.
+- (Nice to have) Implement and rollout delegated publishing to enable updating from multiple devices.
+- (Nice to have) Implement and rollout versioning system so that nodes can see full update history.
+
+##### What does done look like?
+_What specific deliverables should completed to consider this project done?_
+
+Publishing IPNS record takes just a few miliseconds. Shutting laptop and loading `https://ipfs.io/ipns/{key}` week later on other device loads the content published as fast as loading corresponding `https://ipfs.io/ipfs/{cid}`.
```
lidel, line 99: I'd like to apply below change, to make it extra clear the goal here is to solve IPNS rot: 

```suggestion
- Publishing IPNS record takes just a few miliseconds. 
- Shutting laptop and loading `https://ipfs.io/ipns/{key}` week later on other device loads the content published as fast as loading corresponding `https://ipfs.io/ipfs/{cid}`.
- It is possible to pin IPNS address to keep it around and ensure it resolves even when original publisher is gone (AKA solving IPNS rot).
  - Node follows IPNS record updates and keeps re-publishing latest one if original author disappears.
  - Node follows content updates behind the IPNS record, and update content pin every time IPNS record changes. 
  - It is possible to opt-out from this behavior to re-publish IPNS record for `en.wikipedia-on-ipfs.org` without pinning all the data.
```

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,168 @@
...
+
+#####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+- People and teams are storing IPNS keys in [DNSLink][] records as opposed to CIDs, because that provides a more effective and simpler way to publish updates.
+- We see people storing IPNS addresses in ENS to save on blockchain transactions costs.
+- We see new projects leveraging IPNS (when human readable name is not a concern) instead of working around it wit [DNSLink][].
```
lidel, line 110: This would have to be normalized to account for overall growth, but should be easy to track by processing gateway logs:
```suggestion
- We see new projects leveraging IPNS (when human readable name is not a concern) instead of working around it with [DNSLink][].
- We see more requests for  `/ipns/{libp2p-key}` paths on our public gateway (dweb.link &  ipfs.io).
```

proposals/Reliable Mutability Primitive.md
```diff
@@ -0,0 +1,168 @@
...
+##### Dependencies/prerequisites
+<!--List any other projects that are dependencies/prerequisites for this project that is being pitched.-->
+
+- Functioning IPNS in browsers
+- Enabling pubsub in go-ipfs
+
+##### Future opportunities
+<!--What future projects/opportunities could this project enable?-->
+
+- [Petname][] system for memorable but non canonical names (kind of how our phones allow us to not remember numbers).
```
lidel, line 137: I'd like to add dogfooding IPNS in update mechanisms here:
```suggestion
- [Petname][] system for memorable but non canonical names (kind of how our phones allow us to not remember numbers).
- IPNS-based autoupdate mechanisms for our own stack (eg. go-ipfs, ipfs-desktop etc)
```

lidel (commented):
I removed IPNS from #42 and would like to add some IPNS-specific things  here  (details below)

mikeal (approved):

mikeal:
i’d like a review approval from @aschmahmann on this one

Gozala:
I think it is worth pointing out that this proposal was mostly about:

1. Documenting current state of mutability in IPFS
2. Making a case for addressing this problem

During iterations it became clear that different strategies (with various tradeoffs) could be used to address it. I have proposed #66 as one such strategy and was hoping that other proposals would surface as well. I also attempted to further document current limitations https://hackmd.io/@gozala/state-of-ipns across environments that I challenge other previously discussed strategies to improve IPNS across the board.

mikeal:
@Gozala should we close this PR in favor of #66 then?

Gozala:
> @Gozala should we close this PR in favor of #66 then?

Not sure. I feel like current proposal structure does not support decoupling problem statement that attempts to clarify why we should address the problem, from concrete plan how can we do that. In many cases including this one it would make sense to have single problem statement (like this document) and multiple proposals to address it (which could link to this instead of reiterating).

Currently "Brief plan of attack" attempts to capture what @aschmahmann had been saying on various channels, to whom I'd defer this call.

aschmahmann:
IMO going down the path of this proposal better prepares us for a decentralized solution while also having "exit points" towards #66 if needed.

Briefly my understanding of this proposal is:
- Implement third party IPNS republishing
- go-ipfs default use of PubSub and IPNS over PubSub
- js-ipfs fully implementing IPNS over PubSub (i.e.  [spec](https://github.com/ipfs/specs/blob/master/naming/pubsub.md) which includes the fetch protocol) and using it by default

Briefly my understanding of #66 is:
- Define fast lane name resolution specification.
- Define name keeper service specification.
- Define name routing service specification.
- Implement name routing service in go-ipfs
- Implement name keeper service in go-ipfs
- Implement fast lane name resolution across web, go, node ipfs.
- Deploy name routing service to PL operated boostrap nodes

But what if instead these were almost the same:
- Define fast lane name resolution specification -> ✅ fetch protocol
- Define name keeper service specification -> third party IPNS republishing + "pinning API" to ask a third party to do it for you
- Define name routing service specification -> ✅ DHT provider records
- Implement name routing service in go-ipfs -> ✅  DHT provider records
- Implement name keeper service in go-ipfs -> third party IPNS republishing + "pinning API" to ask a third party to do it for you
- Implement fast lane name resolution across web, go, node ipfs -> IPNS over PubSub (Go ✅,  JS ⚠)
- Deploy name routing service to PL operated boostrap nodes -> ✅ delegated routing

There's a reasonable point made in #66 that maybe IPNS records should have more information in them like routing hints (e.g. where is a good node to start my query), and while we can do this with pretty low effort because of how the DHT processes IPNS records I also don't think it's really going to be necessary once we've got IPNS over PubSub actually working.

There is a risk/issue with the IPNS over PubSub approach wherein PubSub falls over if you try and sign up for too many topics. However, if this becomes an issue we can easily pivot towards #66 by making a new router type that only uses the Fetch protocol instead of also using PubSub. The tradeoff is that we no longer get updates pushed to us and have to poll for updates, but it's an easy pivot to make if we need to.

momack2:
Is this something we can grant to an ecosystem team? third party IPNS
publishing sounds like something that could be executed through an
RFP/devgrant...

aschmahmann:
> Is this something we can grant to an ecosystem team? third party IPNS
publishing sounds like something that could be executed through an
RFP/devgrant...

Maybe? The toughest part of the job is doing the plumbing in go-ipfs which shouldn't be too hard to do but is also not intuitive. I'd be ok giving this a shot as long as we do regular check-ins to check for alignment.

Gozala:
> But what if instead these were almost the same:
> 


I think it is worth considering:

1. Complexity between approach proposed in #66 and below. Broadly speaking #66 attempts network querying by caching name keeper addresses forever at:

   1. Local level
   2. Name routing level

    And removes need for disturbing a network when publishing updates by making it point to point operation.

    Furthermore by embedding name keeper address in the DNSLink all human readable name resolutions could effectively be point to point operations.

   _It is less decentralized, but but that is a conscious decision to remove all of the overhead in happy path. It does not replace need for fully distributed solution as it is still needed as a slower fallback._

2. Dependencies

#66 intentionally does not specify how name routing service API nor name keeper API nor how things get replicated. That is because I want to be able to make progress on IPNS without blocking on pubsub or DHT in JS or other things. 

> 
>     * Define fast lane name resolution specification -> ✅ fetch protocol
>

Not sure what you're referring to here. Assuming it's name keeper resolution I think [fetch protocol](https://github.com/libp2p/specs/blob/master/fetch/README.md) is probably a good way to go about it. Although protocol seems generic enough that I feel some semantic meaning might be useful.

> 
>     * Define name keeper service specification -> third party IPNS republishing + "pinning API" to ask a third party to do it for you

#66 intentionally makes publishing an operation that does not require disturbing the rest of the network. I think it is a good compromise, maybe it is not ? Worth comparing tradeoffs at least.

> 
>     * Define name routing service specification -> ✅ DHT provider records
>

in #66 intention was to decouple name routing from DHT, pubsub etc... If node is aware of some name keeper(s) it can provide routing. It does not even needs to be IPFS node.

>
> 
>     * Implement name routing service in go-ipfs -> ✅  DHT provider records
>

👍

> 
>     * Implement name keeper service in go-ipfs -> third party IPNS republishing + "pinning API" to ask a third party to do it for you
>

In #66 intenion was that anything, as simple as cloudless lambda could play that role in the system. All it needs is a key value store as opposed to whole IPFS node. 

> 
>     * Implement fast lane name resolution across web, go, node ipfs -> IPNS over PubSub (Go ✅,  JS ⚠)
>

I really want us to arrive to a place where you can build distributed applications on the web without having to have to operate ipfs nodes on the server, because that is where many choose AWS as cheaper and simpler alternative. #66 was geared towards enabling that so might be worth considering.

It is true however that it creates a system in which a name resolution depends on an authoritative node, however remedy is that name owner can freely swap that authority which I think provides a good compromise.

> 
>     * Deploy name routing service to PL operated boostrap nodes -> ✅ delegated routing

Load on routing nodes as per #66 is designed to be a lot smaller than I imagine it would be with delegated routing.

BigLep:
Assigning to grants since it won't get worked on in the short term with the current program structure. @mikeal is going to evaluate grant possibility.

PR #21, opened 2021-02-17
Project Flare aka. decentralised hole punching

aarshkshah1992:
NAT traversal in libp2p via Hole Punching with Limited Relays

### DIFF
--- a/proposals/Hole Punching.md
+++ b/proposals/Hole Punching.md
@@ -8,103 +8,79 @@
 
 # Purpose & Impact
 
+### Background
 
-### <ins>Background</ins>
+Given the pervasiveness of IPV4 peers that are behind NATs on the internet, NAT traversal is an essential requirement for a peer to peer application. The inability to traverse around NATs means that such NATT’d peers are NOT reachable on the network and are thus unable to provide any meaningful service to the network, nor interact with network participants under protocol patterns that require inbound connections (e.g. dialbacks).
 
-Given the pervasiveness of IPV4 peers that are behind NATs on the internet, NAT traversal is an essential requirement for a peer to peer application. The inability to traverse around NATs means that such NATT’d peers are NOT reachable on the network and are thus unable to provide any meaningful service to the network, nor interact with network participants under protocol patterns that require inbound connections (e.g. dialbacks). \
-\
 Libp2p currently executes NAT traversal using[ Circuit Relays](https://docs.libp2p.io/concepts/circuit-relay/) wherein publicly dialable Relay servers relay the entirety of user traffic to peers that are NATT’d. This approach does NOT scale because:
-
-
-
 
 1. It costs bandwidth on the Relay server.
 2. There is NO real incentive to be a Relay server.
 3. Introduces communication latency between the two peers that are interfacing via the Relay server.
 
-\
 A more scalable approach to NAT traversal is to enable direct communication between the peers via a technique called _[Hole Punching](https://en.wikipedia.org/wiki/Hole_punching_(networking))_. Hole punching removes the need to relay _all_ traffic between two peers via a Relay server. Instead, relay servers are used merely at the connection bootstrapping phase, to convey signalling between two peers intending to connect to each other, sufficient to facilitate NAT hole punching. In most cases, such traffic represents a minimal fraction of the full user payload traffic.
 
-Hole punching  **has been shown to have ~60% success for TCP & ~80% success for QUIC**  ([MIT paper on Hole Punching](https://pdos.csail.mit.edu/papers/p2pnat.pdf)). It has been widely studied in academia and has also been widely adopted in real world large scale p2p applications  ([Tailscale blog on NAT traversal](https://tailscale.com/blog/how-nat-traversal-works/)). Pervasive web protocols like WebRTC rely on ICE, which uses signalling (STUN) to facilitate hole punching. \
-\
+Hole punching  **has been shown to have ~60% success for TCP & ~80% success for QUIC**  ([MIT paper on Hole Punching](https://pdos.csail.mit.edu/papers/p2pnat.pdf)). It has been widely studied in academia and has also been widely adopted in real world large scale p2p applications  ([Tailscale blog on NAT traversal](https://tailscale.com/blog/how-nat-traversal-works/)). Pervasive web protocols like WebRTC rely on ICE, which uses signalling (STUN) to facilitate hole punching.
+
 Based on [metrics that we collected](https://protocollabs.grafana.net/d/lNGQTv9Zz/hydra-boosters?orgId=1&refresh=5m&from=1613523541216&to=1613534341216&var-head=All&viewPanel=50), we have discovered that:
-
-
 
 *   **~80% peers in the current DHT network are NOT dialable/reachable**.
 *   Relay servers funded and operated by PL aren’t enough given the size of our network, and are also expensive to run.
 
 We posit that implementing hole punching will enable direct connectivity for a significant portion of the network, and will improve overall QoS of the network by increasing the connectivity. In combination with future efforts to strengthen WebRTC support, it will also enable browser-centric use cases to be reliably built on libp2p.
 
+### Intent
 
-### <ins>Intent</ins>
+By implementing a _[Limited Relay](https://docs.google.com/document/d/1bhoVGitiB2rr6i8cVKvwkQONHvsfBxqcK76ePVW_JVs/edit#heading=h.s248pnlqjmsm)_ protocol that ONLY provides the resources and bandwidth needed to coordinate a hole punch instead of acting as full fledged data transfer Relays, we can get public DHT servers to run the Limited Relay protocol without costing them too much bandwidth/resources and thus ALSO get a pervasive & large scale Hole Punching co-ordination infrastructure that can help scale hole punching to the size of the network.
 
-By implementing a _[Limited Relay](https://docs.google.com/document/d/1bhoVGitiB2rr6i8cVKvwkQONHvsfBxqcK76ePVW_JVs/edit#heading=h.s248pnlqjmsm)_ protocol that ONLY provides the resources and bandwidth needed to coordinate a hole punch instead of acting as full fledged data transfer Relays, we can get public DHT servers to run the Limited Relay protocol without costing them too much bandwidth/resources and thus ALSO get a pervasive & large scale Hole Punching co-ordination infrastructure that can help scale hole punching to the size of the network. \
-\
 Note that Hole Punching fails if either peer is behind what is known as a _[Symmetric NAT](https://dh2i.com/kbs/kbs-2961448-understanding-different-nat-types-and-hole-punching/)_ as opposed to a _[Cone NAT](https://dh2i.com/kbs/kbs-2961448-understanding-different-nat-types-and-hole-punching/)_,  which is why Hole Punching does not deliver 100% success rates, but it’s much better than the alternative which is NO _direct_ connectivity for ALL NATT’d peers.
 
-Also, based on anecdotal evidence in the wild and engineering war stories (see [Tailscales’s blog](https://tailscale.com/blog/how-nat-traversal-works/)), **Cone NATs are much more pervasive in Home ISPs over Symmetric NATs, which justifies the~60-80% success for Hole Punching**. \
-\
+Also, based on anecdotal evidence in the wild and engineering war stories (see [Tailscales’s blog](https://tailscale.com/blog/how-nat-traversal-works/)), **Cone NATs are much more pervasive in Home ISPs over Symmetric NATs, which justifies the~60-80% success for Hole Punching**.
+
 Given that libp2p is a library to build peer to peer applications,**implementing Hole Punching in libp2p will allow any application/network that builds on top of libp2p to also benefit from Hole Punching**. This will be a significant step forward in the ease of building well connected p2p networks and will be a major added motivation for developers to build on top of our stack. Based on our research, **no such library that comes with out of the box NAT traversal via Hole Punching exists out in the wild today and we have the opportunity to provide the first such library & infra to herald the age of better connectivity in Web3 apps**.
 
-
-### <ins>Assumptions & hypotheses</ins>
-
-
+### Assumptions & hypotheses
 
 *   Applications that build on top of our stack want peers to be _directly_ reachable from the network even though they are behind a NAT (~80% peers in the current DHT network).
 *   PL is not willing to keep funding expensive  bandwidth-unrestricted Relay servers as the network keeps growing to enable data transfer to/from NATT’d peers.
 *   Users would love to use our p2p stack if doing so means the applications they build get NAT traversal via Hole Punching out of the box.
 *   Enabling better connectivity in IPFS, Filecoin and other applications that build on top of Libp2p in a world suffering from the tyranny of NATs is a success metric/priority for us.
 
-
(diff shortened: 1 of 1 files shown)


proposals/Hole Punching.md
```diff
@@ -0,0 +1,180 @@
...
+
+Also, based on anecdotal evidence in the wild and engineering war stories (see [Tailscales’s blog](https://tailscale.com/blog/how-nat-traversal-works/)), **Cone NATs are much more pervasive in Home ISPs over Symmetric NATs, which justifies the~60-80% success for Hole Punching**. \
+\
+Given that libp2p is a library to build peer to peer applications,**implementing Hole Punching in libp2p will allow any application/network that builds on top of libp2p to also benefit from Hole Punching**. This will be a significant step forward in the ease of building well connected p2p networks and will be a major added motivation for developers to build on top of our stack. Based on our research, **no such library that comes with out of the box NAT traversal via Hole Punching exists out in the wild today and we have the opportunity to provide the first such library & infra to herald the age of better connectivity in Web3 apps**.
```
alanshaw, line 48: ✊ yes!

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+*   **Phase1**
+    - ~90% hole punching success if both peers are behind a Cone NAT in the first dog-fooding phase using PL hosted Limited Relays.
+*   **Phase 2**
+    - Optimise based on Dogfooding results and metrics and ship the feature using statically configured PL hosted Limited Relays.
+*   **Phase 3**
+    - Once we conclude that PL hosted Limited Relays are stable,  ship a release that turns on the Limited Relay protocol in public DHT servers.
+*   **Phase 4**   
+    - Once ~30% of public DHT servers upgrade to support the Limited Relay protocol (measure using Hydra Boosters), ship automated discovery & use of Limited Relays to coordinate a hole-punch rather than using statically configured Limited Relays servers.
+*   **Phase 5**
+    - ~90% hole punching success if both peers are behind a Cone NAT  in the second dog-fooding phase that uses AutoRelay to discover and connect to Limited Relays in the wild rather than using statically configured Limited Relays.
```
raulk, line 106: Provided some private feedback on this wording, but just to summarise, this roadmap reads like a mixture of goals and tasks. It's not entirely clear what needs to be done for each milestone to be done. Consider wording these as task lists, or definition of done (deliverables), but not as a mixture of abstract goals and high-level tasks.

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Success criteria
+
+*   Dog-fooding phases deliver ~90% success for labbers using Home ISPs with Cone NATs (TCP & QUIC).
+*   No bugs related to Hole Punching failures if both peers involved in a Hole Punch are behind a Cone NAT (we have good PRs for and will ship code/tools for users to detect their NAT type).
+*   Users do not file bug reports about their public DHT peers getting DDosed/consuming too much bandwidth/resources because of acting as Limited Relays.
+*   We receive great traction and feedback on the ease of use and robustness of Hole Punching on channels such as Twitter, user surveys and from our community of users/partners.
```
raulk, line 113: We should fold the success criteria into each milestone.

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+
+Also, based on anecdotal evidence in the wild and engineering war stories (see [Tailscales’s blog](https://tailscale.com/blog/how-nat-traversal-works/)), **Cone NATs are much more pervasive in Home ISPs over Symmetric NATs, which justifies the~60-80% success for Hole Punching**.
+
+Given that libp2p is a library to build peer to peer applications,**implementing Hole Punching in libp2p will allow any application/network that builds on top of libp2p to also benefit from Hole Punching**. This will be a significant step forward in the ease of building well connected p2p networks and will be a major added motivation for developers to build on top of our stack. Based on our research, **no such library that comes with out of the box NAT traversal via Hole Punching exists out in the wild today and we have the opportunity to provide the first such library & infra to herald the age of better connectivity in Web3 apps**.
+
```
Gozala, line 41: Hyperswarm does https://pfrazee.hashbase.io/blog/hyperswarm

https://github.com/hyperswarm/hyperswarm
https://github.com/hyperswarm/discovery#dholepunchpeer-cb

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+The User builds an application such as IPFS, Filecoin, Ethereum etc. on top of Libp2p and gets _direct_ connectivity between NATT’d peers (albeit ~60% for TCP & ~80% for QUIC) out of the box without bringing up additional infrastructure and with minimal configuration.
+
+### Impact
+🔥🔥🔥
+
+*   Libp2p will be one of the first libraries in the web3 ecosystem that provides Hole Punching and hence better connectivity out of the box. This will increase the functionality of our stack, and will encourage more developers to build on top of it.
+*   PL’s applications such as Filecoin & IPFS will get turbocharged with better connectivity as they build on top of libp2p.
+*   Important projects such as Eth2, 0x etc that build on top of libp2p will ALSO get this huge benefit should they choose to use it. Eth2 is likely to need this feature in Phase 2, which introduces browser-based light clients.
+*   New browser-centric use cases will be possible when this functionality is implemented in js-libp2p.
```
Gozala, line 59: Could this be elaborated a bit more, I am not sure I understand what are those use cases or how hole punching would make browser nodes dialable.

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+
+**Summary: The idea of a peer to peer library that makes NAT traversal via Hole Punching easy and pervasive is a very important & exciting development in the world of peer to peer applications.**
```
Gozala, line 61: I would go as far as claim that there is a general assumption that p2p networking library would come with Nat traversal & Hole Punching built-in.

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Leverage
+
+_How much would nailing this project improve our knowledge and ability to execute future projects?_
+
+🎯🎯
```
Gozala, line 67: I think higher score is warranted here

```suggestion
🎯🎯🎯
```

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+*   **Phase 5**
+    - ~90% hole punching success if both peers are behind a Cone NAT  in the second dog-fooding phase that uses AutoRelay to discover and connect to Limited Relays in the wild rather than using statically configured Limited Relays.
+
+### Success criteria
+
+*   Dog-fooding phases deliver ~90% success for labbers using Home ISPs with Cone NATs (TCP & QUIC).
+*   No bugs related to Hole Punching failures if both peers involved in a Hole Punch are behind a Cone NAT (we have good PRs for and will ship code/tools for users to detect their NAT type).
+*   Users do not file bug reports about their public DHT peers getting DDosed/consuming too much bandwidth/resources because of acting as Limited Relays.
+*   We receive great traction and feedback on the ease of use and robustness of Hole Punching on channels such as Twitter, user surveys and from our community of users/partners.
+
```
Gozala, line 114: I would suggest to add a criteria such as "IPFS node is able to tell if it can be dialed (hole punched to), combined with some troubleshooting interface that can guide a node operator in terms of what to do to make node reachable"

proposals/Hole Punching.md
aarshkshah1992: @Gozala Does HyperSwarm use decentralised signalling ? How does it signal/co-ordinate hole punching ?

proposals/Hole Punching.md
aarshkshah1992: @Gozala 

Given that JS peers can't connect to go peers directly right now, shipping the Limited Relay based decentralised signalling is the first step in implementing a WebRTC transport in go & js libp2p where signalling does NOT rely on centralised STAR servers.

vasco-santos: "JS peers can't connect to go peers directly right now"

The root reason for this is that `go-ipfs` does not have the `websockets` transport by default. Also these nodes are not usually reachable with a SSL+DNS multiaddr, which is essential for the browser.

Unless we plan to have `webRTC` used by default in `go-ipfs`, the browser limitations will not be solved.

I remember that Chrome had speed limitations for `webRTC`, which made websockets a faster option. @Gozala do you know if this is still the case? If this is the case, I think it would be better to focus on a solution for generating certificates instead of `webRTC` in go. (and of course enable websockets by default in go)

Gozala: > I remember that Chrome had speed limitations for `webRTC`, which made websockets a faster option. @Gozala do you know if this is still the case? If this is the case, I think it would be better to focus on a solution for generating certificates instead of `webRTC` in go. (and of course enable websockets by default in go)

I do not know what speed limitations chrome or other browsers put on WebRTC, but there are problems with WebRTC beyond that:

1. WebRTC is not implemented in workers threads.
   - This makes building responsive apps a lot more challenging given all the other worker that is happening on main thread
   - Prevents IPFS node sharing a node across contexts like tabs or iframes.
2. I have anecdotal reports that in practice WebRTC is impractical without TURN servers, suggesting that often times data is relayed anyway.
3. Again anecdotal, evidence suggests that WebRTC seems to cause significant CPU load

All the above combined often leads teams to pivot towards WebSocket based solutions. I also heard teams reporting reduced bills when operating WebSocket based relay as opposed to TURN servers.

Please take all this with a grain of last, because I have not seen any comprehensive studies to support anecdotal evidence.

proposals/Hole Punching.md
Gozala: I'm afraid I do not know how it goes about it. Reading the description of it (quoted below) makes me think that bootstrap nodes facilitate this.

> If your IP and port is consistent across the bootstrap nodes holepunching usually works.

We could try to ask @mafintosh

proposals/Hole Punching.md
Gozala: And here is some data regarding ICE success rate https://medium.com/the-making-of-whereby/so-i-read-that-20-of-webrtc-calls-fail-67b185e49765

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+- @vyzo
+- @aarshkshah1992
+- @raulk
+
+# Purpose & Impact
+
+### Background
+
+Given the pervasiveness of IPV4 peers that are behind NATs on the internet, NAT traversal is an essential requirement for a peer to peer application. The inability to traverse around NATs means that such NATT’d peers are NOT reachable on the network and are thus unable to provide any meaningful service to the network, nor interact with network participants under protocol patterns that require inbound connections (e.g. dialbacks).
```
marten-seemann, line 13:
```suggestion
Given the pervasiveness of IPv4 peers that are behind NATs on the internet, NAT traversal is an essential requirement for a peer to peer application. The inability to traverse NATs means that such NATT’d peers are NOT reachable on the network and are thus unable to provide any meaningful service to the network, nor interact with network participants under protocol patterns that require inbound connections (e.g. dialbacks).
```

marten-seemann (approved):

mikeal:
There are still some unaddressed comments from @raulk

aschmahmann:
@momack2 @raulk I noticed that this project was moved into the Completed tab. Is it done? i.e. what needs to happen for me to ship a release of go-ipfs where a peer behind a NAT can pull data from another peer behind a NAT?

mxinden:
> @momack2 @raulk I noticed that this project was moved into the Completed tab. Is it done? i.e. what needs to happen for me to ship a release of go-ipfs where a peer behind a NAT can pull data from another peer behind a NAT?

Adding to the above, does this project include writing specifications? If I am not mistaken specs are missing for [_circuit relay v2_](https://github.com/libp2p/go-libp2p-circuit/pull/125), the [_signaling protocol_](https://github.com/libp2p/go-libp2p/pull/1057) (though there is https://github.com/libp2p/specs/pull/173) and [_AutoNAT_](https://github.com/libp2p/specs/issues/180).

jacobheun:
Tracking issue in libp2p https://github.com/libp2p/go-libp2p/issues/1039

M3T4F34R (approved):

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+
+Given that libp2p is a library to build peer to peer applications,**implementing Hole Punching in libp2p will allow any application/network that builds on top of libp2p to also benefit from Hole Punching**. This will be a significant step forward in the ease of building well connected p2p networks and will be a major added motivation for developers to build on top of our stack. Based on our research, **no such library that comes with out of the box NAT traversal via Hole Punching exists out in the wild today and we have the opportunity to provide the first such library & infra to herald the age of better connectivity in Web3 apps**.
+
+### Assumptions & hypotheses
+
+*   Applications that build on top of our stack want peers to be _directly_ reachable from the network even though they are behind a NAT (~80% peers in the current DHT network).
+*   PL is not willing to keep funding expensive  bandwidth-unrestricted Relay servers as the network keeps growing to enable data transfer to/from NATT’d peers.
+*   Users would love to use our p2p stack if doing so means the applications they build get NAT traversal via Hole Punching out of the box.
```
nothingismagick, line 46: Hope we are not too late to the party, but can definitely agree from the perspective of downstream consumers of libp2p-rust, that not only is hole-punchy NAT traversal the ultimate goal, but the point above about the cost of running full relays also means that, since there is no incentive to do so, such holistic systems are likely to suffer from centralisation concerns.

https://github.com/iotaledger/stronghold.rs/pull/210

proposals/Hole Punching.md
nothingismagick: True. We were surprised when it didn't come delivered "out of the box". In fact, initially we built a publicly reachable "mailbox" solution that leveraged polling. Obviously terrible at scale for both transit and storage.

proposals/Hole Punching.md
```diff
@@ -0,0 +1,146 @@
...
+
+*   We are able to remove static configuration of  PL hosted Limited Relays in lieu of discovering Limited Relays. And better still, completely shut them down without causing disruptions to the Hole Punching abilities of the network (disruptions can be measured using Github issues, bug reports and PL hosted Cone NATT’d nodes that scrap the network for unreachable peers with Cone NATs, attempt Hole Punching with them and report corresponding metrics).
+
+### Counterpoints & pre-mortem
+
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+### Alternatives
+
+Maybe implementing a WebRTC transport in go-libp2p that performs signalling/co-ordination via Limited Relay servers can help solve the connectivity problems that hole punching seeks to address but that means that we get tied to using WebRTC as a transport. Compared to that, implementing hole punching as a first class feature in Libp2p makes the whole feature transport agnostic.
```
nothingismagick, line 125: Just to raise another point here, webrtc is not available in all webviews, and even though there is promising work being done on the rust front with eg https://webrtc.rs - projects that build atop of Tauri are very likely to prefer to use a wss connection. 

Similarly, there are privacy concerns since not everyone will run their own STUN services and then probably use something like a public google service.

See: https://github.com/tauri-apps/wry/issues/85

proposals/Hole Punching.md
nothingismagick: If I recall correctly, it uses webrtc ICE/STUN.

bam80:
Ping?

PR #22, opened 2021-02-17
Support Large IPLD/IPFS DAGs

Stebalien:


proposals/large-ipld-dags.md
```diff
@@ -0,0 +1,116 @@
...
+_What must be true for this project to matter?_
+
+There is no easy way (e.g., no out-of-band deals) to store large (> sector size) IPLD DAGs while
+preserving the DAG structure.
+
+#### User workflow example
+_How would a developer or user use this new capability?_
+
+* `lotus client deal` should accept an IPLD selector.
+* `lotus cleint deal` should automatically split large DAGs between multiple sectors.
```
vasco-santos, line 36:
```suggestion
* `lotus client deal` should automatically split large DAGs between multiple sectors.
```

proposals/large-ipld-dags.md
```diff
@@ -0,0 +1,116 @@
...
+All three of the above commands have been implemented.
+
+NOTE: stopping anywhere along the way will yield a useful result. As long as the first step is finished (selector support for `lotus client deal`) the 
+
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+1. Developers can easily store large directory trees on Filecoin.
+2. Developers can easily retrieve individual files from large datasets on Filecoin.
+
```
lidel, line 81: Archive like  [wikipedia_en_all_maxi_2021-02.zim](https://download.kiwix.org/zim/wikipedia/wikipedia_en_all_maxi_2021-02.zim)  is ~80GB, we unpack it and put unixfs DAG on IPFS together with original ZIM, so the final DAG  is 4 or 5 times the original size (ignoring deduplication).
Sounds like a good test case to use for validating deliverable here:

```suggestion
3. Snapshots of English Wikipedia can be stored on Filecoin.
```

proposals/large-ipld-dags.md
```diff
@@ -0,0 +1,116 @@
...
+* `lotus client retrieve` should support retrieving IPLD selectors (dag subsets).
+
+#### Impact
+_How directly important is the outcome to web3 dev stack product-market fit?_
+
+🔥
+
+At the moment, any tool wishing to support storing IPFS files/directories larger than 32GiB will need to store these IPFS files/directories as "raw blocks", throwing away all the DAG structural information. This will make future retrieval deals for subsets of this data infeasible and will make IPFS interop extremely difficult.
+
+This is only one 🔥 because there are plenty of useful sub-32GiB datasets and non-IPFS datasets.
```
aschmahmann, line 46: This is true although there is additional impact here which is enabling people to store compositions of data sets.

If deals already exist on Filecoin for a dataset and then someone wants to reference that dataset (or some part of it) within theirs then the data has to be duplicated and stored in two separate deals. With this feature as long as there is a way to discover mappings of CID -> miner with CID (currently out of band, but is a required part of retrieval market work) then users don't need to store the same data twice (or worry about compositions exceeding 32GiB)

proposals/large-ipld-dags.md
```diff
@@ -0,0 +1,116 @@
...
+
+The primary risk is that there may be a lack of demand to store large IPFS-formatted datasets in Filecoin. That is, users storing large datasets (> 32GiB) may all be using custom formats and may not care about IPFS files/directories, partial retrieval, etc.
+
+Another risk is that the IPLD selector language may be insufficient to describe useful selectors over IPFS data. It should be at least possible to _store_ DAG subsets using IPLD selectors, but we may need new selectors to, e.g., download individual IPFS files.
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+1. Don't support datasets > 32GiB.
+2. Store large datasets as raw objects instead of IPFS files and accept the fact that these datasets
```
aschmahmann, line 94: I guess a 3rd one like this could include allowing users to send a parallel DAG structure that only contains links if they want to be queryable and accepting that our selector options will be limited and some dealing with this manifest may be a pain

rvagg:
Title could do with some work. "in Filecoin" would be helpful, but this is about partial dags too, so maybe it's "Support Storage and Retrieval of Large & Arbitrary IPLD DAGs in Filecoin".

But maybe this is up to three separate projects:

 * Support arbitrarily large DAGs in Filecoin
 * Support arbitrary and incomplete DAGs natively in Filecoin
 * Selector support for retrieval of partial DAGs

Stebalien:
Yeah, this could be split into 3 mini projects. But the overarching goal is to be able to support large datasets, both for storage and retrieval. I'm not sure what breaking it into three parts would give us.

proposals/large-ipld-dags.md
```diff
@@ -0,0 +1,117 @@
...
+   will be difficult to query/retrieve from IPFS.
+
+#### Dependencies/prerequisites
+<!--List any other projects that are dependencies/prerequisites for this project that is being pitched.-->
+
+None.
```
willscott, line 100: we should think about if this can be deferred or done in parallel with having the lotus client / market work using ipld-prime

proposals/large-ipld-dags.md
```diff
@@ -0,0 +1,117 @@
...
+_Success means impact. How will we know we did the right thing?_
+
+1. Developers can easily store large directory trees on Filecoin.
+2. Developers can easily retrieve individual files from large datasets on Filecoin.
+3. Snapshots of English Wikipedia can be stored on Filecoin.
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+The primary risk is that there may be a lack of demand to store large IPFS-formatted datasets in Filecoin. That is, users storing large datasets (> 32GiB) may all be using custom formats and may not care about IPFS files/directories, partial retrieval, etc.
```
momack2, line 86: is this the required path for partial retrievability - or is that a somewhat orthogonal (if related) problem?

proposals/large-ipld-dags.md
Stebalien: Just using ipld-prime doesn't get us much. I need to be able to (a) make a deal over a selector and (b) retrieve a selector.

proposals/large-ipld-dags.md
Stebalien: I'm not sure how the comment relates to the paragraph so I may be misinterpreting it.

Step 2 of the "plan of attack" is required for partial retrieval.

proposals/large-ipld-dags.md
warpfork: https://github.com/protocol/web3-dev-team/pull/27 is probably a dependency.

momack2:
@Kubuxu could you review this please?

proposals/large-ipld-dags.md
```diff
@@ -0,0 +1,117 @@
...
+into a sector when serialized as a CAR, it must be converted to raw-blocks, chunked, and then stored
+as those chunks.
+
+Unfortunately:
+
+1. This workaround erases the underlying DAG structure. This makes it difficult to transfer this
+   data for both storage and retrieval. This is especially true when interacting with IPFS.
+2. This workaround requires storing an "overlay" DAG in Filecoin (paying for that storage).
+
+Second, it should be possible to retrieve subsets of DAGs. While the underlying protocols support
```
rvagg, line 22: The protocols support this - I think this is referring to graphsync and the other IPLD pieces down to the data storage - but the CLI doesn't. What about the miner side of this? The wording of this suggests that it's just the client CLI that's blocked on this, is that true? Can an alternative retrieval client use the protocols today to retrieve an arbitrary sub-DAG from a miner or is there more to be done on that side too?

ribasushi: I tested selector-based retrievals way back in August ( using a hardcoded selector in the client directly ) - they worked, in the context of everything else being flaky.

It's not a CLI issue, rather we do not have a decent selector interchange format in general ( a gob of cbor is not something to use over API/CLI )

In other words:
- if today I want to specify a cid - I usually get to do the funny `{ "/":"baf..." }` thing
- if today I want to express a selector - I do... ❓

rvagg: ahhhhh back to the "selector syntax" problem, we should just solve that properly eh? so close https://github.com/ipld/specs/pull/239

PR #25, opened 2021-02-17
Proposal: IPLD in IPFS

mvdan:
Ported from the draft at https://docs.google.com/document/d/1lfJ1ElwkkDfzhN5liBMFywnC7hA1CJu2g-IT1jgdtJw/edit.

### DIFF
--- a/proposals/25-ipld-in-ipfs.md
+++ b/proposals/25-ipld-in-ipfs.md
@@ -71,11 +71,10 @@ This a project whose greatest impact may be unblocking other projects. Among the
 
 #### Confidence
 
-On the scale: 1 -- Low
+Low - Medium (1-3)
 
-Impact: 1 -- We have anecdotal examples of users who are frustrated with their inability to put semantic data into IPFS (todo: collect examples from Eric Myhre, Adin Schmahmann)
-
-Impact 3: Top user request: https://github.com/ipfs/go-ipfs/issues/7909
+* We have anecdotal examples of users who are frustrated with their inability to put semantic data into IPFS
+* It is a recurring user request: https://github.com/ipfs/go-ipfs/issues/7909
 
 ## Project definition
 #### Brief plan of attack
@@ -83,22 +82,26 @@ Impact 3: Top user request: https://github.com/ipfs/go-ipfs/issues/7909
 go-IPFS primarily uses legacy IPLD libraries -- go-ipld-format -- in the context of working primarily with UnixFS v1.
 Use of IPLD primarily flows through the [`DAGService`](https://github.com/ipfs/go-ipld-format/blob/master/merkledag.go), which sits on the BlockService, which is hardcoded to fetch through Bitswap, among other things.
 
-Several things sit on top of `DAGService` -- most prominently UnixFS.
+[Several things](https://hackmd.io/KnC0cZGPRvSQG0g8Z9OMXw) sit on top of `DAGService` -- most prominently UnixFS.
 
 The first goal here is to get to a go-ipld-prime based `DAGService`, as close in interface to the existing DAGService, so that libraries dependent on this service would require only minimal changes.
 
-A first step is proposing an ipld-prime interface for this revamp’d DAG service, and to write a minimal scaffold translating from that interface to the current implementation.
+A first step is proposing ipld-prime interfaces for this revamp’d DAG service, and to write a minimal scaffold translating from that interface to the current implementation.
 
 At that point, two people will continue working within the `dag service -> block service -> bitswap` and plumb through IPLD in that direction, while others will add more IPLD commands to the IPFS command line and handle propagating it to the various top level interfaces.
 
-We can also decide if we want to pursue updating some services that sit on top of `DAGService`.
+There is also a triage for which services that sit on top of `DAGService` are updated to what extent.
+
+There will be work on updating IPFS and IPLD documentation to reflect these changes and describe the new interface.
 
 #### What does done look like?
 
 * P0: I can build IPFS without go-ipld-format
+  * P1: Deprecate go-ipld-format
 * P1: I can import structured data (following links of generic codecs) directly from the IPFS command line
 * P1: structured data is importable and can be queried over the IPFS HTTP API 
 * P2: I can query structured data directly from the IPFS command line
+* P1: Documentation allows plugin authors to update to the new interfaces
 
 ####  What does success look like?
 
--- a/proposals/25-ipld-in-ipfs.md
+++ b/proposals/25-ipld-in-ipfs.md
@@ -2,7 +2,7 @@
 
 Authors: @hannahhoward, template edit by @mvdan
 
-Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+Initial PR: https://github.com/protocol/web3-dev-team/pull/25
 
 ## Purpose &amp; impact 
 #### Background &amp; intent
@@ -81,7 +81,7 @@ Impact 3: Top user request: https://github.com/ipfs/go-ipfs/issues/7909
 #### Brief plan of attack
 
 go-IPFS primarily uses legacy IPLD libraries -- go-ipld-format -- in the context of working primarily with UnixFS v1.
-Use of IPLD primarily flows through the [DAGService](https://github.com/ipfs/go-ipld-format/blob/master/merkledag.go), which sits on the BlockService, which is hardcoded to fetch through Bitswap, among other things.
+Use of IPLD primarily flows through the [`DAGService`](https://github.com/ipfs/go-ipld-format/blob/master/merkledag.go), which sits on the BlockService, which is hardcoded to fetch through Bitswap, among other things.
 
 Several things sit on top of `DAGService` -- most prominently UnixFS.
 
--- a/proposals/25-ipld-in-ipfs.md
+++ b/proposals/25-ipld-in-ipfs.md
@@ -1,4 +1,4 @@
-# [outcome or objective here] 
+# Integrate IPLD Prime into go-ipfs
 
 Authors: @hannahhoward, template edit by @mvdan
 
--- a/proposals/25-ipld-in-ipfs.md
+++ b/proposals/25-ipld-in-ipfs.md
@@ -96,8 +96,9 @@ There will be work on updating IPFS and IPLD documentation to reflect these chan
 
 #### What does done look like?
 
-* P0: I can build IPFS without go-ipld-format
-  * P1: Deprecate go-ipld-format
+* P0: I can develop in IPFS without go-ipld-format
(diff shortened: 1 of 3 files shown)


proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
...
+
+Large; 6-9 weeks.
+
+#### Roles / skills needed
+
+* Initial 3-weeks: 1-2 developers with IPFS & IPLD knowledge creating the ipld-prime DAG Service and translation layer
+* Second 3-weeks: 3-4 developers,
+  one group (1-2 devs) plumbing down (integrating in bitswap and go-graphsync),
+  one group (1-3 devs) plumbing up (moving UnixFS, IPFS command lines over to new interfaces
+* Remaining time: add new IPFS commands to work with structured data
```
willscott, line 148: I'm feeling like this is scope-able to 4 weeks given that we already have a prototype shim that we allocated 3 weeks to.

mvdan: You mean the entire thing, or just this last third item?

willscott: i think it's somewhat up to us to prioritize how many apis and what api surface to expand. We can think about that as a "if we have another 1-2 weeks what's the next most important things" and tack those on, and that'll be more clear as we've enumerated that places that ipld data is hitting up to the boundaries.

willscott: i'm honestly less concerned about CLI commands than i am about api surfaces when IPFS is used in an embedded or library form because that's where pre-existing structured data is more likely to be passing in & out

proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
+# [outcome or objective here]
```
lidel, line 1: I feel this is what the proposal is about?
```suggestion
# Integrate IPLD Prime into go-ipfs
```

proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
...
+* Faster data transfer in IPFS (we can integrate bitswap and graphsync to speed up transfer, we cannot in the current stack)
+* Unlocking greater capabilities in the stack through use of selectors, including to-be-written selectors
+* IPLD based databases and graph querying as optimization mechanism
+* Standardizing IPLD across the stack including Filecoin (see meta doc above)
+* Coherent IPFS and Filecoin retrieval patterns (also needs Filecoin retrieval improvements)
+* Almost any effort to support dApps that work with non-static or non-file data
+
+#### Confidence
+
+On the scale: 1 -- Low
```
BigLep, line 74: Do we have any color as to why we have low confidence for this?

proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
...
+
+Impact: 1 -- We have anecdotal examples of users who are frustrated with their inability to put semantic data into IPFS (todo: collect examples from Eric Myhre, Adin Schmahmann)
+
+Impact 3: Top user request: https://github.com/ipfs/go-ipfs/issues/7909
```
BigLep, line 78: We're listing 2 impact numbers.  Is that intentional?  Also, should this be up in the Impact section above?

proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
...
+
+A first step is proposing an ipld-prime interface for this revamp’d DAG service, and to write a minimal scaffold translating from that interface to the current implementation.
+
+At that point, two people will continue working within the `dag service -> block service -> bitswap` and plumb through IPLD in that direction, while others will add more IPLD commands to the IPFS command line and handle propagating it to the various top level interfaces.
+
+We can also decide if we want to pursue updating some services that sit on top of `DAGService`.
+
+#### What does done look like?
+
+* P0: I can build IPFS without go-ipld-format
```
BigLep, line 98: Do this also allow us to "deprecate" go-ilpd-format?

proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
...
+* P1: I can import structured data (following links of generic codecs) directly from the IPFS command line
+* P1: structured data is importable and can be queried over the IPFS HTTP API 
+* P2: I can query structured data directly from the IPFS command line
```
BigLep, line 101: Why wouldn't this be a P1 as well?  Without this, wouldn't it mean we can import structured data but can't query it?

proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
...
+* We may discover design issues in go-ipld-prime that require significant additional work to make it truly useful
+
+#### Alternatives
+
+* It’s possible we can just provide utilities in IPFS to import structured data and serialize to flat files?
+* It’s possible we can unlock some (thought not all) of the blocked projects without porting go-ipfs to go-ipld-prime
+
+#### Dependencies/prerequisites
+
+None, for the time being
```
BigLep, line 128: While not a hard blocker, should we use this as an opportunity to shore up ipld documentation?  I'm assuming lots of questions are going to come to @warpfork about ipld-prime.  Ideally all his answers should be responses back with links to public documentation (that was previously written or just written).  I thought I heard him say that there has been some IPLD documentation setup/structuring that has impaired this.  @warpfork : do you have any comments?

proposals/25-ipld-in-ipfs.md
```diff
@@ -0,0 +1,148 @@
...
+
+* We can continue to develop and expand data querying capabilities with improved selectors, and an effort to serialize selectors for command lines
+* We can move go-ipld-prime into the Filecoin stack and VM and reach IPLD standardization across the stack
+* See all blocked projects
+
+## Required resources
+
+#### Effort estimate
+
+Large; 6-9 weeks.
```
BigLep, line 140: These numbers are ambiguous.  Are we talking about calendar time or number of dev weeks of effort?

I would scrap this as I think the breakdown below is better as it specifies the number of dev weeks for each phase and how parallelizable it is.  

Aside:  I think it can be helpful to phrase estimates like: (Fictitious example) "This project requires an estimated 20 weeks of dev effort by engineers experienced with the language/technology.  It can be parallelized with up to 4 engineers, thus having a calendar time of 5-20 weeks."

proposals/25-ipld-in-ipfs.md
BigLep: Understood if this isn't known yet, but what does it look like to do this project right/fully?  I'm obviously new and don't have all the context about tradeoffs, but in general, I want to avoid having any asterisks/caveats about this kind of implementation.  I'm saying this for two reasons:
1. Without knowing more, it seems like an Inter-Planetary Stack flagship project like IPFS should showcase/adopt what our premier data structure (IPLD) can do.  
2. We don't know when we're going to pick this thread up again, so I would bias to finishing it off while the context is fresh.  
(but again, please disregard if this is off base)

BigLep (commented):
Thanks for putting this together.  I'm hopeful/excited to see this come to fruition.

proposals/25-ipld-in-ipfs.md
rvagg: I think you could say the tension goes something like this:

1. There's a *ton* of power in the IPLD stack as a full-functioning data system, as outlined here.
2. Developers have just got used to throwing all their data into raw blocks and wrapping structured data around DAG-PB / UnixFS and _just making it work_.

We've been pushing IPLD to do all sorts of powerful things ([e.g.](https://github.com/mikeal/ipsql)), and so have some of our more progressive [ecosystem partners](https://github.com/textileio/go-threads/tree/master/cbor). But perhaps these are just advanced users, and the average web3 dev is perfectly happy _treating everything as a file_, after all, that works perfectly well in unix!

hannahhoward: This was based on the doc that @anorth shared - https://medium.com/@nimay/inside-product-introduction-to-feature-priority-using-ice-impact-confidence-ease-and-gist-5180434e5b15 - which suggested internal assumptions, intuitions, and anecdotal evidence, did not amount to a high number. In general, we have very little data beyond this for almost any of the proposed projects.

proposals/25-ipld-in-ipfs.md
hannahhoward: That's still a bit of a stretch. Perhaps?

proposals/25-ipld-in-ipfs.md
willscott: I think this was a conservative number based on the scale that we haven't done market research, which was indicated as a threshold for a 3 or above. we have heard things depending on this as a user ask from many existing users though, so arguably could push it harder.

proposals/25-ipld-in-ipfs.md
willscott: I think the iteration we heard most recently from @anorth is that the impact score is subjective enough that we might just want the prose without trying to assign a score. I think the 2 paragraphs were different options proposed while drafting this that both ended up in the document.

proposals/25-ipld-in-ipfs.md
willscott: that's the hope!

proposals/25-ipld-in-ipfs.md
hannahhoward: In fairness these are more prioritizations against the time allotted, which is generally how I think of prioritization -- the priority is defined by its relative priorization to other tasks. If we get the P0 done, then we do the P1, then the P2. That's probably not the best way to go about it.

willscott: Having it accessible programmatically in structured ways is probably the implicit p1 while passing in and out of the inherently serialized command line is perhaps less used by _developers_ building on IPFS?

proposals/25-ipld-in-ipfs.md
hannahhoward: So, currently the project's P0 is 'compile go-ipfs without go-ipld-format' -- and that's a pretty big undertaking. The shim is just the beginning. We have to move over all the client libraries which is the big task. I'm also generally not a fan of updating estimates based on some very early proof-of-concept work appearing to go well, as getting to production grade implementation is usually what takes time.

BigLep:
@willscott : I assume you'll ping related Stewards when the "progress report" is ready for folks to look at so they can ask questions while the context is fresh.  My apologies that already happened and I missed it.

BigLep:
Specific actions are being tracked in: https://github.com/ipfs/go-ipfs/pull/7976

mvdan:
I'm assuming that this serves no purpose in being open at this point; the files are always available in this PR even when closed.

PR #27, opened 2021-02-17
Proposal: Selector resource budgets

warpfork:
tl;dr Selectors currently don't have a resource budgeting system, and this means accepting user-defined selectors is a DoS vector.  Work is needed to fix this in order to make Selectors usable in more scenarios.  These scenarios include things wanted in Filecoin implementations, to my understanding.


lidel:
I suspect this would be valuable outside Filecoin context as well. 
AFAIK there is a lot of hidden effort duplication across pinning services like Pinata: they need to manually code some safeguards to protect themselves against DoS via maliciously crafted DAGs.

proposals/SelectorResourceBudgets.md
```diff
@@ -0,0 +1,207 @@
...
+
+One: See remarks about budgeting in the Assumptions & Hypothesis section.
+Some system of resource currency could be associated with this problem as part of the solution.
+(This doesn't necessarily remove the need for engineering work on the Selector system to support it, though,
+which means this should probably be considered a stretch goal or future work rather than an alternative.)
+
+Two: It's possible to work around this in some cases by building APIs around selectors,
+but then only accept a known, pre-specified set of selectors.
+(If I understand correctly, this is how several pieces of Filecoin currently around around this issue.)
+This is not a general workaround, though, and ruins most of the point of Selectors -- they're *supposed* to be user-specifiable.
```
BigLep, line 151: I don't know the specific, usecase but I assume a gimped selector syntax could still be useful potentially for some users depending on their needs.

proposals/SelectorResourceBudgets.md
```diff
@@ -0,0 +1,207 @@
...
+
+- A system that halts but returns insufficient information about why could be frustrating to users,
+  even if it successfully addresses the DoS problem.
+
+- Keep in mind: this proposal only describes implementing this in golang.
+  We do not currently have a javascript Selectors implementation (and creating one is a larger task).
+  This is not a problem per se; it's just something to remember when considering what can be immediately built upon this work.
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
```
BigLep, line 141: I assume, but is there any concept of pagination/partial results that can be applied for selectors?

BigLep (commented):
Thanks for putting this together @warpfork.  Interesting!  Per comments, before engaging deeper on this one, I think it would be helpful to have some more specific usecases that customers are asking for.  Let me know if help is needed to gather any of this.

proposals/SelectorResourceBudgets.md
```diff
@@ -0,0 +1,207 @@
...
+----------------
+
+#### Background & intent
+_Describe the desired state of the world after this project? Why does that matter?_
+
+The status quo is: we have Selectors, and they can be used to describe walks of graphs of data.
+(They're sorta like regexps for DAGs, if that's a useful comparison for you.)
+We want to expose these
+
+The problem is: if a service wants to accept Selectors which are user-specified,
```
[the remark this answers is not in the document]
warpfork: For go-ipfs: right, I could care less if someone uses a local API to hose _themselves_.  But people seem to want to expose these things publicly.  For example, https://github.com/protocol/web3-dev-team/pull/1 seems to imply we're going to have APIs oriented around Selector queries, and does not say much about not letting these be exposed remotely.  This is representative of most conversations I've ever overheard about Selectors and what people want to do with them.  (So, if nothing else, this proposal needed to be made to track the situation!)

For filecoin: I taaaag.... @magik6k ?  (I have repeatedly heard this is wanted, that's the depth limit of my knowing.)

In general: it seems like it's almost a law of human nature that people want to ask arbitrarily complex questions without concern for the costs on the answerer :laughing: / :cry:  The Selectors system seems to be no exception, heh.

proposals/SelectorResourceBudgets.md
warpfork: Then we would need to invent such a gimped syntax.

That's probably _harder_ than planning and implementing budgets within the current syntax.

My general experience with this topic is: you cannot, no matter how big of a gavel you wave and how energetically you wave it, convince people to stop asking for features that would make a system accidentally turing-complete (and thus an unbounded DoS vector).  (This is doubly true when it comes to tree or graph processing, which, in a sudden flash hindsight that only occurs to me fully now, probably ought not be a surprise.)  Therefore: monotonically decreasing budgets, often aka "gas", is the only real way to unambiguously communicate the problem, and thus the only real practical way to solve it.

warpfork: The other route is "invent a gimped syntax ***and*** a compiler that verifies its non-TC-ness" -- and that's possible; that's what eBPF is, if I understand correctly -- but it's a _huge_ amount of work...

... and I sorta wouldn't bet very favorably on if that approach would work for tree/DAG processing scenarios anyway.  I'd rather bet money that it would end up with people wanting to apply the eBPF-like thing repeatedly on every block they visit.

Which would get us back to approximately the same problem with Selectors right now: since those things would "restart" their budget on every block visited, we'd need some... bigger, holistic, monotonically decreasing budget.

proposals/SelectorResourceBudgets.md
warpfork: It has been discussed, but no such thing is implemented nor shipped at present.

Resuming that discussion would probably be a part of the work that would go on while engaging on this project.

warpfork: In practice, it's my understanding that at present, systems using Selectors get in the habit of launching small queries (depth limited, or, constructed to favor left-leaning trees for example) to get started with exploring the data, and use more queries subsequently.

This "works" but obviously leaves some load to the brain of the human crafting the Selector, which isn't really the most desired outcome.

warpfork: Some prior discussion involved ideas like "what if we could ask the selector to return every (N % 3 == 1) blocks?" and similar ideas.  The aim there was to end up with something that you could imagine a system generating automatically in order to fan out queries for data to multiple peers and start getting different fractions of the data back from them in parallel.

This only got to the discussion phase.  There may be neat ideas here, but they trend towards getting complicated, so we pushed them out of the first round of Selector work.

mikeal (approved):

proposals/SelectorResourceBudgets.md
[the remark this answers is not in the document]
warpfork: Big use-case I can't believe I forgot in the earlier comment: *graphsync*.

We seem to talk about the intention to use graphsync between untrusted peers who might be exchanging data without a fee mechanism.  If that's true, then it will be important for such peers to have at least some cost estimation mechanism and cutoff options.

PR #28, opened 2021-02-17
Proposal: End to End Tests for Deal Flow in Lotus

alanshaw:


proposals/28-deal-flow-e2e-tests.md
```diff
@@ -0,0 +1,192 @@
...
+**Ease 8** an estimate that this will take 3 weeks to get up and running. The time is reltively short because there's no changes to Lotus that need to be made, it can be created as a separate greenfield project.
+
+## Project definition
+#### Brief plan of attack
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+1. Identify test scenarios.
+    * See the "Lotus Release Checklist" for some desired E2E tests to include.
+1. Identify tests that will generate benchmarking data.
+1. Decide on runner (BYO or testground?).
```
BigLep, line 113: Does PL have standard "canary" patterns and infrastructure?  For example, in the AWS world I know folks use https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Synthetics_Canaries.html (or roll their own with Lambda).

For TestGround, my total-new-guy understanding is that there is work to be done on test ground itself so it can used within CI.  From Stewards conversations I was listening in on, it sounded like FileCoin has done some of this work, but hasn't fully got it across the line.  Anyways, this sounds like a good area to engage the Stewards on.

proposals/28-deal-flow-e2e-tests.md
```diff
@@ -0,0 +1,192 @@
...
+1. Over time, performance for the collected metrics improves.
+
+#### Counterpoints &amp; pre-mortem
+<!--
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+-->
+
+If the deal flow changes before the project is complete then there will be additional work to do to update the tests either at the end or mid development.
+
+If the tests are "flakey" then their results will not be trusted and they may become abandoned.
```
BigLep, line 153: Agreed.  Should an exit criteria be that the tests can pass X times (e.g., 100) with a pass rate of Y (e.g., 99)?

BigLep (commented):
Thanks for writing this - good stuff!  A couple of things:
1. Long term ownership - who is responsible for the infrastructure to run these tests, keeping them updated, and fixing them if they get flaky (not due to Lotus itself being flaky).
2. What network(s) will these run against?  It would be ideal to have these also run against the production network so we detect network issues before clients report them but understood that is likely out of scope right now and is debatable even if it's worth doing.

proposals/28-deal-flow-e2e-tests.md
```diff
@@ -0,0 +1,192 @@
...
+_How would a developer or user use this new capability?_
+(short paragraph)
+-->
+
+_To be confirmed_. In general it should be runable from CI, and users should be able to initiate ad-hoc runs with particular branches to test changes.
+
+Some ideas:
+
+* Run on every PR to Lotus to identify regressions.
+* Run nightly as a separate project to collect and graph data such as:
```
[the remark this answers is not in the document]
willscott: there is a pl-wide prometheus / grafana instance visible at https://protocollabs.grafana.net/login that would be a natural place for these sorts of metrics to pipe into

proposals/28-deal-flow-e2e-tests.md
```diff
@@ -0,0 +1,192 @@
...
+<!--
+_How would a developer or user use this new capability?_
+(short paragraph)
+-->
+
+_To be confirmed_. In general it should be runable from CI, and users should be able to initiate ad-hoc runs with particular branches to test changes.
+
+Some ideas:
+
+* Run on every PR to Lotus to identify regressions.
```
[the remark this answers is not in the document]
alanshaw: Agree, I've seen a lot of flakey tests for distributed systems. These need to be watertight.

proposals/28-deal-flow-e2e-tests.md
```diff
@@ -0,0 +1,192 @@
...
+* Run nightly as a separate project to collect and graph data such as:
+    * Average "time to deal".
```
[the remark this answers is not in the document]
alanshaw: Yes, if we collect the raw numbers we can organise them into whatever format we want to see. Grafana will help with that :)

proposals/28-deal-flow-e2e-tests.md
```diff
@@ -0,0 +1,192 @@
...
+    * Minimum number of commands to store data.
```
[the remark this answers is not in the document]
alanshaw: Yes, invaluable for identifying the bottlenecks. I'll update the PR to be explicit about collecting individiual command times as well as total time.

proposals/28-deal-flow-e2e-tests.md
alanshaw: I'm somewhat against this because it can mask real issues. In my experience it is almost always because of a flakey test or bug in the code that a test fails, not for other reasons. We want to fix both of those two things, not ignore them.

When there's a legitimate need for this I've seen test frameworks provide a e.g. `retries(3)` function where the test can fail up to 2 times.

proposals/28-deal-flow-e2e-tests.md
alanshaw: I don't know enough about testground to comment authoratatively. My understanding is that a cluster would be made available that would be used for this testing. Hopefully we can orchestrate/trigger from CI or do some work to make that possible.

Maybe @nonsense or @raulk can speak to this? 🙏

alanshaw:
> 1. Long term ownership - who is responsible for the infrastructure to run these tests, keeping them updated, and fixing them if they get flaky (not due to Lotus itself being flaky).

Good question. I'd like to know the story for current testground infra to really be able to comment on that. I assume this is more infra for the infra team but it would be nice if there were an alternative to that.

IMHO if you change the CLI commands for the deal flow then you are responsible for updating the tests. Like with any tests.

Changes to the code that cause a test to periodically fail need to be investigated. That's a PITA because it's not always possible to determine which developer is responsible. It could be the dev who wrote the test, could be the dev who made the change, or could be a combination of the two (the change has caused an assumption in the test to not hold anymore).

I would suggest that it's the responsibility of the sub-team involved in markets/deals to own any disputes but any developer what causes the tests to fail where the reason is obviously because of a change they made should be responsible for fixing them.

This is a good point because it needs buy in from the teams involved. If the tests are not seen as valuable then I'd rather not spend the time. However it is the responsibility of the implementer to make them as reliable and useful as possible.

> 2. What network(s) will these run against? It would be ideal to have these also run against the production network so we detect network issues before clients report them but understood that is likely out of scope right now and is debatable even if it's worth doing.

I'd anticipated for a network to be spun up just for running the tests, but a testnet that is already running might speed up the round trip time.

I assume we'd create actors as needed for running the tests, so I'd expect to receive similar results running the same tests on mainnet...and running on mainnet would be way slower as we'd probably speed up the epoch period for testing. I also don't think we should pollute mainnet with test actors and deals.

proposals/28-deal-flow-e2e-tests.md
nonsense: @BigLep @alanshaw here is what we have built on Testground in terms of canaries and end-to-end deals tests:

1. Testground is deployed at Testground as a Service (TaaS) at https://ci.testground.ipfs.team/

2. Automated tests are run on every merge to `master` from https://github.com/filecoin-project/lotus . [Relevant part of .circleci/config.yaml](https://github.com/filecoin-project/lotus/blob/master/.circleci/config.yml#L296:L325) . Unfortunately these tests are rather flaky, but definitely worth review in the realm of "end-to-end tests for deal flow in Lotus" - one of the testplans is doing exactly that. It is based on `TestDealFlow` test in Lotus.

If developers want to give Testground as a Service a try, I can explain the relevant configs so that you trigger test plan on the shared cluster from your local machine - Lotus testplans are rather heavy in terms of memory requirements at the moment, so it might be difficult to run multiple miners/clients on a laptop.

---

Feel free to reach out to me about more questions related to Testground or these test plans!

dirkmc: Also worth noting there are some end-to-end deal tests in lotus itself:
https://github.com/filecoin-project/lotus/blob/master/api/test/deals.go

mikeal (approved):

BigLep:
Currently waiting on resourcing and after @raulk's Lotus testing sprint.

willscott:
noting that some of this may overlap with the dealbot now doing regular testing of the deal flow

PR #30, opened 2021-02-17
[WIP] Proposal: DHT in JS for effective content/service discovery

jacobheun:
Sharing early, need more time to finish this up.

### DIFF
--- a/proposals/dht-js.md
+++ b/proposals/dht-js.md
@@ -1,0 +1,106 @@
+# Effective content and service discovery in JS
+
+Authors: @jacobheun
+
+Initial PR: TBD
+
+## Purpose &amp; impact
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives.
+-->
+
+**Current State**
+- JS projects rely on delegate and preload nodes to be able to interact with the live network
+- PL hosted delegate/preload nodes are often overloaded negatively impact performance of projects
+
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+TODO
+
+#### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+TODO
+
+#### Impact
+TODO
+
+<!--
+Explain why you have chosen this rating
+What awesome potential impact/outcomes/results will we see if we nail this project?
+-->
+
+#### Leverage
+TODO
+
+<!-- Explain the opportunity or leverage point for our subsequent velocity/impact (e.g. by speeding up development, enabling more contributors, etc)
+-->
+
+#### Confidence
+_How sure are we that this impact would be realized? Label from [this scale](https://medium.com/@nimay/inside-product-introduction-to-feature-priority-using-ice-impact-confidence-ease-and-gist-5180434e5b15)_.
+
+<!--Explain why this rating-->
+
+
+## Project definition
+#### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+The focus of this effort is to build the DHT to function solely in Node.js. Running a DHT in browser is not currently viable.
+
+- Week 0 - Complete the DHT specification (DHT protocol expertise)
+- Create a test plan in Testground to evaluate query performance
+- Implement DHT routing table construction and refresh
+- Implement the improved query logic
+- Implement Peer Eviction logic
+
+
+**Assumptions**
+- Routing table diversity is not criticial and can be implemented later
+- We can start with "Client Mode" only support to minimize the surface area of the solution
+
+#### What does done look like?
+
+- js-ipfs in Node.js ships with the DHT enabled in client mode by default
+- js DHT query times are comparable to Go over TCP/Noise connections
+
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+#### Dependencies/prerequisites
+<!--List any other projects that are dependencies/prerequisites for this project that is being pitched.-->
+
+#### Future opportunities
+<!--What future projects/opportunities could this project enable?-->
+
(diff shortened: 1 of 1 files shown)


proposals/dht-js.md
```diff
@@ -0,0 +1,106 @@
...
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives.
+-->
+
+**Current State**
+- JS projects rely on delegate and preload nodes to be able to interact with the live network
+- PL hosted delegate/preload nodes are often overloaded negatively impact performance of projects
+
```
vasco-santos, line 17:
```suggestion
- We need to recommend teams with JS expertise to learn Golang and use go-ipfs, in order to  overcome the current content routing limitations, which might make these teams look for other alternatives

```

proposals/dht-js.md
```diff
@@ -0,0 +1,106 @@
...
+<!--Explain why this rating-->
+
+
+## Project definition
+#### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+The focus of this effort is to build the DHT to function solely in Node.js. Running a DHT in browser is not currently viable.
```
ZenGround0, line 55: Here's a stupid question from someone with little exposure to js ipfs or js ecosystem at large.  What is the benefit to building out functionality in node.js as opposed to relying on a delegate go-ipfs nodes?  Does the value come from having multiple fully functional implementations?  Is there a positive impact on in-browser nodes from delegating DHT functionality to a node.js ipfs node instead of a go-ipfs node?

proposals/dht-js.md
```diff
@@ -0,0 +1,106 @@
...
+
+**Current State**
+- JS projects rely on delegate and preload nodes to be able to interact with the live network
+- PL hosted delegate/preload nodes are often overloaded negatively impact performance of projects
+
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+TODO
```
vasco-santos, line 22: - A large slice of the developers in the web3 space are JS developers

proposals/dht-js.md
jacobheun: Not a stupid question, and both are viable options with pros/cons and I think there is merit in both. I will expand more here when I have time to flush this out, but I think there is significant work that needs to be done improving remote api access to Go nodes and that should be a separate proposal. There is a lot to be gained there for new developers. The DHT in JS is relatively short term in comparison and immediately improves usability of the JS ecosystem of projects for existing developers (we currently have users leveraging JS and Go in hacky ways to get around some limitations in these systems; flexible IPLD in JS, performant DHT in Go).

proposals/dht-js.md
```diff
@@ -0,0 +1,106 @@
...
+## Purpose &amp; impact
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives.
+-->
+
+**Current State**
+- JS projects rely on delegate and preload nodes to be able to interact with the live network
+- PL hosted delegate/preload nodes are often overloaded negatively impact performance of projects
```
momack2, line 16: @gmasgras this sounds like something that is soluble from an infra perspective. Can you comment on our current preload node uptime/utilization?

proposals/dht-js.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Purpose &amp; impact
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives.
+-->
+
+**Current State**
+- JS projects rely on delegate and preload nodes to be able to interact with the live network
```
achingbrain, line 15: "JS projects" here is not just the browser, it's node, electron, react native etc.  E.g. web, desktop and mobile.

proposals/dht-js.md
achingbrain, line 16: The requirement to preload content also makes JS projects very bandwidth heavy - every piece of data I add to a JS node is transferred to a preload node, this is very expensive in terms of transfer.

achingbrain, line 16: The preload nodes also garbage collect the content periodically so over time the content becomes inaccessible

achingbrain: Using DHT delegates suffers from similar problems, except it's sort of worse - we re-publish every hour so content survives the timed garbage collection, which means re-uploading every block, which scales poorly.

aschmahmann: I'm not sure I understand why delegated routing should be slow/expensive or what preload nodes are necessary in non-browser use cases.

DHT publishing frequency with delegated routing should match what you'd expect with non-delegated routing, no?

aschmahmann: Sorry if I'm just totally missing the idea here, but why are preload nodes and garbage collection on them relevant in this discussion? Aren't they only really needed if a node is unreachable (e.g. in a browser)?

achingbrain: The point above is that the PL hosted infra is often overloaded which is a risk to project functionality.

I was making the additional point that even if the nodes were not overloaded, they garbage collect anything uploaded to them, so the IPFS magic of 'add a file to your node, cat it from another node' is time-limited if the only way that content makes it to another node is via a preload node.

achingbrain: Yes, the frequency should be the same, but as currently implemented with delegated routing, if a block has been garbage collected, you end up sending the data to the server each time for it to create the provider record - the whole DAG, not just the root block since the delegate has to be able to supply everything on behalf of your node.

> what preload nodes are necessary in non-browser use cases

They're not, at least, they shouldn't be - if the DHT implementation was complete and performant, which it isn't.  This proposal is to finish up the work we did durning the hack week and make it so.  Then we could turn off preload/delegate for everything that isn't a browser node.

aschmahmann: With proper delegated routing shouldn't I be making my own provider records that point at me instead of having ones that points at the preload nodes?

vasco-santos: > E.g. I ask the delegated router for the 20 closest peers then directly send them provider record puts.

Yes, but this is a limitation in the browser, as we will not be able to dial most of the network

aschmahmann: Sure, but my comment was about non-browser nodes not needing to store their data on preload nodes.

Browser nodes have a very rough time talking to the rest of the network at the moment, so they need to do a lot more delegating of work. Does anything in this proposal really help the browser node situation?

aschmahmann: I'm confused, this doc states.

> The focus of this effort is to build the DHT to function solely in Node.js. Running a DHT in browser is not currently viable.

Putting data in preload nodes is not required to get NodeJS to play well with the rest of the network.

PR #35, opened 2021-02-17
Filecoin PubSub Observatory project pitch

vyzo:


raulk:
Likely to be folded into #13.

proposals/filecoin-pubsub-observatory.md
```diff
@@ -0,0 +1,82 @@
...
+Pretty good.
+
+
+## Project definition
+#### Brief plan of attack
+
+1. To evolve the current pubsub traces collector (which receives
+   traces from 1.5k concurrent Lotus nodes) to ship the data stream to
+   our logz.io Elasticsearch, where it can be queried with Kibana to
+   explore and build dashboards.
```
willscott, line 49: cc @lanzafame  as well as he thinks about the moving parts here. adding this stream shouldn't be bad.

@vyzo - can you clarify if the data stream at the trace collector is semi-structured logs or structured data of observed peers / scores?

vyzo: It's fully structured, we collect protobufs (and can also emit json in the trace collector).
Here is the pb definition: https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/trace.proto

For the scores and statistics we don't currently do any collection, this will have to be added to lotus (or possibly a side process).

mikeal:
Did this get folded into #13? Can we close?

PR #37, opened 2021-02-17
[PLACEHOLDER] A better Lotus client-side library

raulk:
Items identified: Improving usability of client-side API library: A lightweight Lotus client API import, so that code depending on Lotus doesn’t have to build FFI and other weirdness; better documentation for the API; add support for API versioning; API dev console (OpenRPC or Swagger defs?).

### DIFF
--- a/proposals/xx-better-lotus-client-side-library.md
+++ b/proposals/xx-better-lotus-client-side-library.md
@@ -1,0 +1,99 @@
+# A better Lotus client-side library
+
+Authors: <!-- List authors' GitHub or other handles -->
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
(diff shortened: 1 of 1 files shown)


magik6k:
Note that the lotus API doesn't depend on filecoin-ffi - we even have a test for that: https://github.com/filecoin-project/lotus/blob/master/api/api_test.go#L28

OpenRPC support is almost done, we just need to push it over the finish line - https://github.com/filecoin-project/lotus/pull/4711

Support for API versioning shouldn't be too hard as most pieces were designed to support it at some point, and it would unblock tons of things, like native multisig support in message-sending apis

ribasushi:
@magik6k I just happened to extract [as a standalone app](https://github.com/filecoin-project/slingshot-stats/blob/722003809/main.go) something yesterday that is super-trivial: retrieve/decode the list of deals from the state. For that [I **had** to use ffi stubs](https://github.com/filecoin-project/slingshot-stats/blob/722003809/go.mod#L17), there was no way around it

PR #39, opened 2021-02-17
proposals: added automated-go-ipfs-releases.md

aschmahmann:
Coming from some comments made at a recent stewards meeting. Very WIP, also not clear if this belongs in this repo or not.

cc @Stebalien

### DIFF
--- a/proposals/automated-go-ipfs-releases.md
+++ b/proposals/automated-go-ipfs-releases.md
@@ -1,0 +1,121 @@
+# [outcome or objective here] 
+
+Authors: @aschmahmann
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
+A proposal should contain enough detail for others to understand how this project contributes to our team’s mission of product-market fit
+for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on,
+and any other information relevant for prioritizing this project against others.
+It does not need to describe the work in much detail. Most technical design and planning would take place after a proposal is adopted.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). 
+Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+Currently the process of performing a go-ipfs release is tedious and is a barrier to having smaller more frequent releases. Having a more automated release process would allow us to iterate faster.
+
+Additionally, it would be great if our release process was such that we felt comfortable having an option for go-ipfs to automatically update on start instead of making users update. This would also be a step towards removing user barriers with package managers like `snap` which are not CLI tool friendly.
+
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+- Doing go-ipfs releases manually must impede release cycle time in a way that impacts project velocity, or
+- The way in which users currently update go-ipfs is sufficiently an impediment that either it causes users to have a bad experience with the stack (e.g. by running into bugs which have long been fixed), or it significantly harms development velocity by needing to take into account old non-upgraded nodes
+
+#### User workflow example
+_How would a developer or user use this new capability?_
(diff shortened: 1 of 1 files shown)


proposals/automated-go-ipfs-releases.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Project definition
+#### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+#### What does done look like?
+_What specific deliverables should completed to consider this project done?_
+
+It takes less than 1hr to do a go-ipfs release from when the release is ready to be tagged, and users can automatically update to the latest version of go-ipfs.
```
lidel, line 81: We may need to use.. IPFS :trollface:  for automatic updates to work in China:

> fwupd, the Linux firmware update service used by Lenovo, Google, Acer, Dell and more, now supports IPFS. Specifically, it was added to improve performance of firmware update transfers out of China. https://blogs.gnome.org/hughsie/2021/02/16/fwupd-1-5-6/

proposals/automated-go-ipfs-releases.md
```diff
@@ -0,0 +1,121 @@
+# [outcome or objective here]
```
BigLep, line 1: Update?

proposals/automated-go-ipfs-releases.md
```diff
@@ -0,0 +1,121 @@
...
+-->
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+Currently the process of performing a go-ipfs release is tedious and is a barrier to having smaller more frequent releases. Having a more automated release process would allow us to iterate faster.
```
BigLep, line 30: Can we quantify this a bit?
1. How much dev effort?
2. How much calendar time?
3. How many manual steps?
4. How many unique individuals need to be involved?

We can then compare the current state for these metrics to the proposed state somewhere in the doc.

proposals/automated-go-ipfs-releases.md
```diff
@@ -0,0 +1,121 @@
...
+
+Additionally, it would be great if our release process was such that we felt comfortable having an option for go-ipfs to automatically update on start instead of making users update. This would also be a step towards removing user barriers with package managers like `snap` which are not CLI tool friendly.
```
BigLep, line 32: This seems really important.  Does it belong in a separate proposal?

proposals/automated-go-ipfs-releases.md
```diff
@@ -0,0 +1,121 @@
...
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+- Doing go-ipfs releases manually must impede release cycle time in a way that impacts project velocity, or
+- The way in which users currently update go-ipfs is sufficiently an impediment that either it causes users to have a bad experience with the stack (e.g. by running into bugs which have long been fixed), or it significantly harms development velocity by needing to take into account old non-upgraded nodes
+
+#### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+A go-ipfs maintainer could run a script or GitHub action that would automatically perform a release from a given branch instead of manually performing the release.
```
BigLep, line 44: I'm wondering if this should expand.  I'm wondering something like the following as an example:

Once a commit is pushed to the release branch, in addition to our tests, it automatically builds/deploys to PL infrastructure in a controlled way (e.g., small region on day 1 (catch glaring bugs), big region on day 2 (catch issues that only show up under load), remaining regions on day 3).  If at any point our alarms trip, we have autorollback, the deployment pipeline blocks, and we get notified during business hours about the bug to fix.  

I think the value of this would be that we would have high confidence on whenever we want to cut/market a new release version as it has been baking on PL infrastructure in production already and if there is a production issue we can quickly pinpoint the change that caused it.

proposals/automated-go-ipfs-releases.md
BigLep: I assume this is 1 person 1 hr?  I assume Bifront/infra time is separate?

Also, what goes into getting a "release ready to be tagged"?  Is that manual effort, if so should this proposal have that in scope so we drive that down?

proposals/automated-go-ipfs-releases.md
```diff
@@ -0,0 +1,121 @@
...
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+- There are external factors slowing down releases other than the process itself
```
BigLep, line 93: Do we know of any of these?

proposals/automated-go-ipfs-releases.md
```diff
@@ -0,0 +1,121 @@
...
+- It takes too much effort to automate the release process
```
BigLep, line 94: If we have an idea on the current and future state plus an estimate of how much work this will take, we can assess the return on investment.  For example, estimated data that would be help:
1. Amount of dev effort this process takes today
2. Amount of dev effort the process will take after doing this project
3. Estimated effort to do this project
4. Number of releases we do in a year

BigLep (commented):
Thanks for writing this up!

PR #42, opened 2021-02-17
Proposal: Reliable resolution of human-readable names

lidel:
- Main focus here is improving the story behind human-readable names
  - extremely impactful when it comes to  user onboarding
  - we have some gaps which limit the usefulness of DNSLink in the context of alternatives to DNS system managed by ICANN
- There is some overlap with https://github.com/protocol/web3-dev-team/pull/19  around what needs to be fixed in IPNS, and if tackled separately could decrease the scope of this project.

cc @Gozala @autonome @aschmahmann

### DIFF
--- a/proposals/41-human-readable-mutable-names.md
+++ b/proposals/41-human-readable-mutable-names.md
@@ -1,0 +1,225 @@
+# Reliable and extensible resolution of human-readable mutable names
+
+Authors: @lidel
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/41 <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+## Purpose &amp; impact 
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+We see CID, IPNS record, and DNSLink being de-facto standard for cryptographically-verifiable content-addressing and human-readable names on the distributed web. 
+
+There are three distinct types of addresses:
+  - immutable content:  `/ipfs/{cid}`
+  - mutable content
+      - cryptographic, IPNS record signed with a specific libp2p key: `/ipns/{libp2p-key-cid}` 
+      - human-readable, [DNSLink](https://dnslink.io): `/ipns/{dns-name}`
+        (which in turn, resolves to either an immutable or mutable content path from DNS TXT record)
+
+
+**There are known gaps in the mutable story:**
+- IPNS records are only published by the original author, which severely limits the utility of our mutable names
+    - It is not possible to pin/republish `/ipns/{libp2p-key-cid}` names by other peers, so when original publisher disappears, IPNS links no longer resolve and we see "IPNS rot". 
+    - **Due to IPNS rot people simply don't use it:**
+        - Instead publishing updates via `ipfs name publish {cid}`, website operators are forced to constantly update DNSLink's DNS TXT record to point at reliable immutable paths(`dnslink=/ipfs/{cid}`). 
+            - It works around the need for running a service responsible for constant republishing of IPNS record.
+            - This is especially bad for security-sensitive use cases where human-readable names are not required. Effectively means DNS is used in contexts where cryptographic addressing should be used instead.
+- **OS-level DNS resolver is used for all DNS TXT record lookups.** 
+  - IPFS node is unable to control which DNS resolver is used for specific TLD, all queries go to global DNS server configured in the operating system
+    - This harms competition and forces picking "winners" prematurely. For example, if user changes their DNS resolver to one that is capable of resolving all TLDs over ENS, they are unable to resolve TLDs from OpenNIC or UnstoppableDomains. 
+  - DNS queries are most likely sent in plain text and can be spoofed in LAN or MITMd at the ISP level. At the same time DNS over HTTPS is not supported, and it is not trivial for regular user to set up a custom DNS server to change this behavior.
+
+**TLDR:** **We need to fix and future-proof the way mutable names work in our stack:**
+- Make IPNS useful on its own, and inside of DNSLink records.
+    - DNS TXT record should be set only once, updates should be published over IPNS.
+- Future-proof the way we do human readable names by
+    - leveraging DNS protocol (not the DNS network run by ICANN) for interop with existing and future user agents and naming systems
+    - allowing flexible configuration to improve security (DoH) and interop (DNSLink)
+
+
+
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+- Browser is the  main distribution channel for dapps.
+- Dapps are simply a subset of all static websites loaded from IPFS.
+- Most of websites hosted on IPFS have DNSLink set up for human-readable name.
+- We have mainstream browsers like Opera and [Brave](https://brave.com/ipfs-support/) ship support for `ipfs://` and `ipns://` URIs. 
+    - Human readable names like `ipns://en.wikipedia-on-ipfs.org` work thanks to DNSLink
+    - This removes huge UX friction for dapp developers. 
+    - Vendors like Brave are committed to [surfacing DNSLink supportin the UI](https://github.com/brave/brave-browser/issues/13609) which will improve onboarding even further.
+- Alternatives to [ICANN](https://en.wikipedia.org/wiki/ICANN)  exist, and we want them to work with our stack to decrease dependency on a single organization.
+    - [ENS](https://ens.domains) provides custom TLDs (`.eth`) and leverage DNSLink for interop with IPFS and delegated lookups with clients that are unable to run own blockchain resolver.
+        - Updating `ipfs-ns` `contenthash` is a chore and costs extra (gas etc). 
+        - Using `ipns-ns` means setting `contenthash` only once, introduces cost savings and simplifies publishing. 
+    - We see multiple actors in the space providing either own TLDs or attempting to replace ICANN as the top-level authority: https://unstoppabledomains.com , http://opennic.org, https://handshake.org etc.
+    - We don't want to pick winners too early.
+    - When embedded in user agent (Brave) we want to follow user choices regarding DNS resolution.
+    
+
+
+
+#### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+- Improving human-readable naming removes huge DX/UX friction for dapp users and developers. 
+    - Messaging collapses to "Install Brave or Opera and open ipns://example.com"
+    - They use IPNS for publishing updates without ever touching DNS/ENS etc beyond initial setup. Most likely save money on gas and/or admin time.
+    - One-liner for setting up a custom DNSLink resolver improves onboarding for naming systems not supported by default and enables vendors like Brave to provide UI for changing resolution settings in more user-friendly manner.
+
+
+#### Impact
+_How would this directly contribute to web3 dev stack product-market fit?_
+
+<!--
+Explain how this addresses known challenges or opportunities.
+What awesome potential impact/outcomes/results will we see if we nail this project?
+-->
+
+- IPNS provides a reliable solution for  cryptographically-verifiable mutable names that won't break, as long someone is re-publishing the latest record.
+    - Just Works (TM)
+    - There is no PKI, no centralization, full transparency and ownership.
+- DNSLink provides interop for human-readable names and enables independence from PKI and ICANN.
+    - Just Works (TM)
+    - No vendor-specific client is included. 
+    - No proprietary APIs or formats. RFC-compliant DNS only.
+    - We promote competition, letting the best solution win.
+    - User agency is respected. 
+
+
+#### Leverage
+_How much would nailing this project improve our knowledge and ability to execute future projects?_
+
+<!--
+Explain the opportunity or leverage point for our subsequent velocity/impact (e.g. by speeding up development, enabling more contributors, etc)
+-->
+
+
+- Brave wants to support alternatives to ICANN  out of the box. 
+    - By having resolver configurable per TLD namespace, we will be able to integrate with their resolution logic, providing seamless experience on `ipns://` out-of-the-box 
+- Every new project can experiment with IPFS via DNSLink integration without our involvement.
+    - We no longer need to say "no" just to keep our code base small or worry about optics of "picking winners and losers".
+- Fixing IPNS rot would enable use of `dnslink=/ipns/{libp2p-key-cid}`
+    - Improved UX because DNS record needs to be set only once
+    - Reduced cost for blockchain-based alternatives to ICANN (like ENS), owner uses IPNS for publishing updates, saving on gas.
+
+#### Confidence
+_How sure are we that this impact would be realized? Label from [this scale](https://medium.com/@nimay/inside-product-introduction-to-feature-priority-using-ice-impact-confidence-ease-and-gist-5180434e5b15)_.
+
+Medium.
+
+<!--Explain why this rating-->
+
+- After five years we see DNSLink being the only viable solution for human-readable names. 
+    - ENS works out of the box on every IPFS Gateway and in Brave thanks to DNSLink interop.
+    - Cloudflare embraced DNSlink years ago, and they took over [the eth.link ENS resolver/gateway](https://blog.cloudflare.com/cloudflare-distributed-web-resolver/).
+    - `ipns://` in address bar of Brave is an extremely powerful visual and a confidence booster about entire stack
+- We see alternatives to ENS popping up, but enabling them requires changes to our code bases and is not a trivial task.
+- Unfortunately, `dnslink=/ipns/` is a rare sight, due to IPNS rot, and people default to DNSLink even in cases where use of cryptographic identifiers would be a better choice.
+
+## Project definition
+#### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+- Tackle IPNS and DNS fixes in parallel, in both Go and JS
+- Introduce configurable DNS resolver per TLD while adding DNS over HTTPS support to entire stack
+
+#### What does done look like?
+_What specific deliverables should completed to consider this project done?_
+
+- DNS over HTTPS is supported in js/go-ipfs/libp2p
+    - User agency around DNS is respected: every code path doing a DNS lookup (resolving DNSLink, `/dnsaddr`, `/dns*` multiaddrs) should use either explicit or implicit DNS resolvers defined in IPFS node configuration file.
+      - By default OS-level DNS resolver is used as catch-all, but the user can delegate resolution of all or specific TLDs to specific DoH resolvers.
+      - Node tries to resolve with the most specific resolver (if present for TLD), and then fall back to the global one.
+    - Visual Aid / config mock-up
+      ```json
+        "DNS": {
+          "Resolvers": {
+            "*.eth": "https://eth.link-or-cloudflare/path/to/dns-query",
+            "*.crypto": "https://unstoppablesomething.example.com/path/to/dns-query",
+            "*.libre": "https://www.opennic.org/path/to/their/dns-query", 
+            "*": "https://mozilla.cloudflare-dns.com/dns-query"
+          }
+        }
+      ```
+- DNSLink provides interop for human-readable names and enables independence from PKI and ICANN.
+    - Anyone can set up and run own naming system without our involvement, blessing or endorsement.
+    - No vendor-specific client is included. 
+- IPNS provides a reliable solution for cryptographically-verifiable mutable names that won't break, as long someone is re-publishing the latest record.
+    - IPNS re-publishing works seamlessly, out of the box, just like providing CIDs for blocks in local datastore (pinned or not).
+    - It is possible to pin IPNS address to keep it around and ensure it resolves even when original publisher is gone.
+        - Node follows IPNS record updates and keeps re-publishing latest one if original author disappears.
+        - Node follows content updates behind the IPNS record, and update content pin every time IPNS record changes. 
+            - It is possible to opt-out from this behavior to re-publish IPNS record for `en.wikipedia-on-ipfs.org` without pinning all the data.
+
+
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+- We see % of `dnslink=/ipns/` going up.
+- We see more IPNS records on DHT
+- We no longer see requests for supporting new naming systems.
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+ 
+- Configurable per TLDs [DNS over HTTPS](https://en.wikipedia.org/wiki/DNS_over_HTTPS) resolver for  may be not enough to mitigate concerns around centralized nature of ICANN's DNS.
+- IPNS may have other problems that block it from practical use in DNSLink records (eg. resolution still taking way longer than immutable path, making initial page load slow).
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+- Deprecate IPNS.
+- Announce ENS being the official solution for DNS names and replace every DNS lookup to go via ENS.
+- Do nothing and pay the opportunity cost when a better alternative to ENS appears but can't be used with DNSLink due to non-ICANN TLDs.
+
+#### Dependencies/prerequisites
+<!--List any other projects that are dependencies/prerequisites for this project that is being pitched.-->
+
+- DNS over HTTPS support in go/js-ipfs and go-/js-libp2p
+- IPNS record republishing/pinning
+
+
+#### Future opportunities
+<!--What future projects/opportunities could this project enable?-->
+
+- Dev grants / collabs with  various naming projects to provide DNSLink gateway (similar to https://eth.link) and DNS over HTTPS endpoint for resolving them securely, including JS in browsers.
+    - Supporting competition, providing venue for rapid adoption and user onboarding, seeing what sticks.
+- P2P swarm resolvers and quorum acceptance criteria 
+    - We should look into a way to harden DNS resolution without introducing any dependency on any complex public PKI (DNSLink should not require DNSSEC).
+    - Instead of delegating DNS lookup to some DoH endpoint, leverage swarm of peers to increase trust in DNS lookup results.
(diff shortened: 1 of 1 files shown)


Gozala:
One thing that is not clear to me here is why IPFS needs to allow plugging alternative name resolution algorithms. Seems like they should be plugging themself at the OS layer instead.

My concern is that primary value of URL is that it provides **universal** way for sender to send a name that receiver can use to obtain content regardless of their configurations. If that becomes conditioned on node configurations of sender and receiver they loose value and in fact can harm the ecosystem by creating a reputation of unreliable or too complicated.

Another thing worth considering is that DIDs seem to be gaining momentum, at least in web3 space, maybe that could be a place to plug alternative naming and resolution schemes as opposed to overloading DNS.

Huge +1 on making IPNS reliable so that DNSLinks use `/ipns/` as opposed to `/ipfs/`

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+            - They point at reliable immutable paths (`dnslink=/ipfs/{cid}`) (no IPNS record means it resolves fast and never break).
+            - It works around the need for running a service responsible for constant republishing of IPNS record.
+            - This is a pragmatic choice, however:
+              - adds admin related to DNS updates (need to learn DNS provider's API for automating this, which introduces soft vendor-lock-in)
+              - most likely not the best choice for security-sensitive use cases where human-readable names are not required: means DNS is used in contexts where cryptographic addressing should be used instead.
+- **OS-level DNS resolver is used for all DNS TXT record lookups.** 
+  - IPFS node is unable to control which DNS resolver is used for specific TLD, all queries go to global DNS server configured in the operating system
+    - This harms competition and slows down innovation in the decentralized naming space. For example, if user changes their DNS resolver to one that is capable of resolving all TLDs over ENS, they are unable to resolve TLDs from OpenNIC or UnstoppableDomains.
```
BigLep, line 42: n00b questions:
1. Is there a DNS resolver that is capable of resolving all TLDs over ENS?  Is this more about something that could come in the future (per https://ccnso.icann.org/sites/default/files/field-attached/presentation-dns-blockchain-ens-24jun19-en.pdf )?
2. Is this a general issue even at the OS level that a user won't be able to have DNS that works with ENS, OpenNIC, and UnstoppableDomains together?  Can someone have a local DNS proxy that can resolve using ENS, OpenNIC, UnstoppableDomains, etc.?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+- **OS-level DNS resolver is used for all DNS TXT record lookups.** 
+  - IPFS node is unable to control which DNS resolver is used for specific TLD, all queries go to global DNS server configured in the operating system
+    - This harms competition and slows down innovation in the decentralized naming space. For example, if user changes their DNS resolver to one that is capable of resolving all TLDs over ENS, they are unable to resolve TLDs from OpenNIC or UnstoppableDomains. 
+  - Given how DNS is implemented in OS and ISPs, queries are most likely sent in plain text and can be spoofed in LAN or MITMd at the ISP level. At the time of writing this proposal DNS over HTTPS is not supported, and it is not trivial for regular user to set up a custom DNS server to change this behavior.
+
+**TLDR:** **We need to fix and future-proof the way mutable names work in our stack:**
+- Make IPNS useful on its own, and inside of DNSLink records.
+    - DNS TXT record should be set only once, updates should be published over IPNS.
```
BigLep, line 47: This is https://github.com/protocol/web3-dev-team/pull/19 right?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+- Dapps are simply a subset of all static websites loaded from IPFS.
+- Most of websites hosted on IPFS have DNSLink set up for human-readable name.
+- We have mainstream browsers like Opera and [Brave](https://brave.com/ipfs-support/) ship support for `ipfs://` and `ipns://` URIs. 
+    - Human readable names like `ipns://en.wikipedia-on-ipfs.org` work thanks to DNSLink
+    - This removes huge UX friction for dapp developers. 
+    - Vendors like Brave are committed to [surfacing DNSLink supportin the UI](https://github.com/brave/brave-browser/issues/13609) which will improve onboarding even further.
+- Alternatives to [ICANN](https://en.wikipedia.org/wiki/ICANN)  exist, and we want them to work with our stack to decrease dependency on a single organization.
+    - [ENS](https://ens.domains) provides custom TLDs (`.eth`) and leverage DNSLink for interop with IPFS and delegated lookups with clients that are unable to run own blockchain resolver.
+        - Updating `ipfs-ns` `contenthash` is a chore and costs extra (gas etc). 
+        - Using `ipns-ns` means setting `contenthash` only once, introduces cost savings and simplifies publishing.
```
BigLep, line 68: The convenience sense.  Question about .ENS users: are they going to be comfortable/supportive of mutating changes happening outside of the blockchain?  Haven't they now effectively lost the transaction history for part of their domain (i.e., the CID for the content the domain is pointing to)?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+    - Vendors like Brave are committed to [surfacing DNSLink supportin the UI](https://github.com/brave/brave-browser/issues/13609) which will improve onboarding even further.
+- Alternatives to [ICANN](https://en.wikipedia.org/wiki/ICANN)  exist, and we want them to work with our stack to decrease dependency on a single organization.
+    - [ENS](https://ens.domains) provides custom TLDs (`.eth`) and leverage DNSLink for interop with IPFS and delegated lookups with clients that are unable to run own blockchain resolver.
+        - Updating `ipfs-ns` `contenthash` is a chore and costs extra (gas etc). 
+        - Using `ipns-ns` means setting `contenthash` only once, introduces cost savings and simplifies publishing. 
+    - We see multiple actors in the space providing either own TLDs or attempting to replace ICANN as the top-level authority: https://unstoppabledomains.com , http://opennic.org, https://handshake.org etc.
+    - We want to enable innovation in the decentralized naming space. This means removing ourselves as gatekeepers of what a valid domain name is.
+    - When embedded in user agent (Brave) we want to follow user choices regarding DNS resolution.
```
BigLep, line 71: Oh interesting.  To make sure I'm following this, using an example like Brave, Brave is going to delegate `ipns/*` to the bundled go-ips binary, which currently will be using the OS DNS resolver rather than any custom DNS resolver set within the browser (e.g., DoH resolver that can be set in browser preferences).  You're saying it would be ideal for Brave to pass its DNS resolution configuration to the go-ipfs binary as well so they can be in sync, right?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+    
+
+
+
+#### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+- Improving human-readable naming removes huge DX/UX friction for dapp users and developers. 
+    - Messaging collapses to "Install Brave or Opera and open ipns://example.com"
```
BigLep, line 81: This is nice in that it doesn't rely on the ICANN network or HTTP gateways (thus working with other DNS networks and/or without HTTP gateways).  That said, the story is already good on the ICANN network right with HTTP gateways (e.g., A user can open Brave and hit ipns://libp2p.io ), right?  Basically, I want to make sure I understand this right, and it seems like the value prop here is for people who care about full decentralization.

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+    - They use IPNS for publishing updates without ever touching DNS/ENS etc beyond initial setup. Most likely save money on gas and/or admin time.
+    - One-liner for setting up a custom DNSLink resolver improves onboarding for naming systems not supported by default and enables vendors like Brave to provide UI for changing resolution settings in more user-friendly manner.
```
BigLep, line 83: > enables vendors like Brave to provide UI for changing resolution settings in more user-friendly manner

I'm not quite following this.  Can you add a bit more detail?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+      ```
+- DNSLink provides interop for human-readable names and enables independence from PKI and ICANN.
+    - Anyone can set up and run own naming system without our involvement, blessing or endorsement.
+    - No vendor-specific client is included. 
+- IPNS provides a reliable solution for cryptographically-verifiable mutable names that won't break, as long someone is re-publishing the latest record.
+    - IPNS re-publishing works seamlessly, out of the box, just like providing CIDs for blocks in local datastore (pinned or not).
+    - It is possible to pin IPNS address to keep it around and ensure it resolves even when original publisher is gone.
+        - Node follows IPNS record updates and keeps re-publishing latest one if original author disappears.
+        - Node follows content updates behind the IPNS record, and update content pin every time IPNS record changes. 
+            - It is possible to opt-out from this behavior to re-publish IPNS record for `en.wikipedia-on-ipfs.org` without pinning all the data.
```
BigLep, line 169: Should this get added to https://github.com/protocol/web3-dev-team/pull/19 ?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+
+
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+- We see % of `dnslink=/ipns/` going up.
```
BigLep, line 179: Where will be able to see this?  Can our gateways detect this and/or is this a metric we can get Brave?  
Should we maybe have the first task in the project plan to be to generate/collect the metric so we can measure our impact?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+- We see more IPNS records on DHT
```
BigLep, line 180: Do we have these metrics today?  Should we add an item to the project plan to generate/collect this?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+- We no longer see requests for supporting new naming systems.
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+ 
+- Configurable per TLDs [DNS over HTTPS](https://en.wikipedia.org/wiki/DNS_over_HTTPS) resolver for  may be not enough to mitigate concerns around centralized nature of ICANN's DNS.
```
BigLep, line 186: Do you have any anecdotes that lead to think this could be the case?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+- IPNS may have other problems that block it from practical use in DNSLink records (eg. resolution still taking way longer than immutable path, making initial page load slow).
```
BigLep, line 187: Do we have any indicating data on what the performance will be like when https://github.com/protocol/web3-dev-team/pull/19 gets done (e.g., IPNS has PubSub)?  I assume @aschmahmann may know?

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+- Deprecate IPNS.
```
BigLep, line 192: n00b question: is it possible to deprecate a protocol like this?  I assume this is work to do the customer outreach/communication.

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,235 @@
...
+- Announce ENS being the official solution for DNS names and replace every DNS lookup to go via ENS.
```
BigLep, line 193: Would customer actually be ok with this given the Gas fees to publish a new CID?

BigLep (commented):
This is awesome - great work!  No expectation to respond to everything in GitHub.  I'm also game to connect verbally if that's easier.

One suggestion: should we scope this proposal to:
1. DNS over HTTPS is supported in js/go-ipfs/libp2p
2. DNSLink provides interop for human-readable names and enables independence from PKI and ICANN.

and let everything about "IPNS provides a reliable solution for cryptographically-verifiable mutable names that won't break, as long someone is re-publishing the latest record" live in https://github.com/protocol/web3-dev-team/pull/19 ?

This would allow this proposal to simplify a bit and I assume they could be executed/delivered in parallel (with the best customer value occurring when both are completed)?

lidel:
@Gozala the ability to override OS-level resolver is crucial for both user choice/security (OS-level one is most likely plaintext from ISP) and for our browser integration story.  Brave wants to add suport for ENS via DoH at https://eth.link/dns-query, with more systems to come,  and I believe go-ipfs should provide primitives to follow DNS resolver choices user made via their  user agent. 

Being able to support non-ICANN names is something we get as a side effect of this work. We already support ENS in a hacky, hardcoded way in go-ipfs. What we propose to do here is to remove vendor-specific code and use  DNS for interop. 

My plan is to ship go-ipfs with some predefined DoH endpoints, so TLDs like `.eth` work out of the box, but the user has the choice to change the default resolver (right now, user has none).

As for DIDs, I don't believe there are any primitives for human-readable names, so out of scope here(?)

proposals/42-human-readable-mutable-names.md
lidel: Ad 1. Yes, https://eth.link provides endpoint at `https://eth.link/dns-query`, but we will use it only for resolving `*.eth`
Ad 2. Correct, my opinion is that requiring users to run DNS proxy is not realistic. Ability to specify DoH resolver per TLD  removes the need for that, but if someone wants to run own DNS resolver, they can also do this  here – just point at localhost one.

proposals/42-human-readable-mutable-names.md
lidel: Yes, this is overlap with fixing IPNS.

proposals/42-human-readable-mutable-names.md
lidel: ENS users have a choice between pointing at IPFS identifiers in one of two namespaces:
- `ipfs-ns` – immutable identifiers – never change
- `ipns-ns` - mutable cryptographic identifiers, only owner of the publishing key can make change and update there it points at

The latter is still a cryptographic identifier, but yes, it is being mutated outside of blockchain. 
For many that is a killer feature, because it means they don't need to pay gas every time they want to update their website. IPNS makes ENS way cheaper, more practical and I'd argue safer – you don't need to take out your hardware wallet to update your blog.

proposals/42-human-readable-mutable-names.md
lidel: Yes, your understanding is correct :+1:

proposals/42-human-readable-mutable-names.md
lidel: Supporting non-ICANN TLDs is a value prop for decentralization crowd, but even ICANN TLDs will benefit from this in privacy/security dimensions:

- virtually all default DNS people have is some plaintext DNS server from ISP
- browsers  are moving towards using custom DoH resolvers instead of the OS one to avoid privacy leaks over unencrypted DNS 
  - if go-ipfs does not do the same, we are privacy leakers: https://github.com/ipfs-shipyard/ipfs-companion/issues/900
- public gateways are running go-ipfs too
  - there is value for gateway operator in being able to change resolver to something that hides DNS queres from their cloud provider: limits privacy leak, and is way harder to spoof, as it is no longer a matter of compromising DNS at the same provider

proposals/42-human-readable-mutable-names.md
lidel: I meant that Brave may provide UI in their settings for changing default DNS over HTTPS resolver endpoint, or one for resolving ENS.

proposals/42-human-readable-mutable-names.md
lidel: Yes, both Gateway and Brave could resolve DNSLink and then store the number of DNSLink requests  + % between mutable and immutable requests.  
But as this is IPNS-specific, probably should go to #19

proposals/42-human-readable-mutable-names.md
lidel: I suspect we could add IPNS record metrics  to engine behind https://dht.ecosystem-dashboard.com
But this is also something for #19

proposals/42-human-readable-mutable-names.md
lidel: Cynical commentary goes like this:
- DoH by definition points at some centralized HTTPS server :trollface: 
- We don't want to ship some preconfigured DoH server for all ICANN names with go-ipfs, to avoid pointing all users at the same endpoint
  - DoH endpoint choice should be made by user or user agent
  - Still, it is possible that most people will use DoH endpoints from Google or Cloudflare, because those are fast and reliable, so we mitigated privacy leak and mitigated ISP-level spoofing, but then put all eggs in one basket

proposals/42-human-readable-mutable-names.md
lidel: iirc ipns-over-pubsub is no longer slowed down by DHT on the initial lookup (as long both resolver and publisher have it enabled), so it should be comparable with immutable paths (even if a bit slower, user won’t notice it because DHT  for resolving final path will take way more)

proposals/42-human-readable-mutable-names.md
lidel: In the current form it  has limited use on the public network.
However, I suspect IPNS most likely works great in private networks where peers are connected to each other + run pubsub.  
We have some ecosystem data from interviews with a subset of our users, understanding their use cases and proposing alternatives (or making it private-swarm-only feature)  would be first step, but I agree, deprecating IPNS is unlikely, given ENS and others already supporting it.

proposals/42-human-readable-mutable-names.md
lidel: Chicken and egg, of sorts: fixing IPNS would remove the problem of gas cost.

However, the way I see it,  the real problem here is not the cost of gas or chore of updating ENS record.
Using this as a global resolver would be just a charade: most domains will remain on ICANN network for years to come, so ENS  would act as _a centralized_ proxy for 99,9....% of ICANN lookups :see_no_evil:

lidel:
@BigLep Yes, we should remove the overlap. I'll try to move IPNS-related parts to #19 tomorrow.

BigLep:
@lidel: one other comment/idea here:  Does the work in this proposal and/or https://github.com/protocol/web3-dev-team/pull/19 help NFT creators/owners in any way?  That is a key usecase we want to ensure the PL stack is supporting in a user-friendly way.  If it is, can you call it out as that will help with prioritization.

lidel:
@BigLep  I refactored this proposal to focus on adding configurable DNS resolvers to our stack + mentioned how improving human-readable names improve user onboarding story for NFT / dapps / static websites. 

IPNS parts are out of the scope of this proposal,   but left them in explainer in the beginning, as it is important to see the full mutability picture. IPNS improvements will be tracked in https://github.com/protocol/web3-dev-team/pull/19, like you suggested.

BigLep:
@lidel: should we also add in that Brave expecting this DoH work to get so that DNS queries are not leaked?   I assume we'd add any links that provide more context about why this is important for this partner specifically.

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,258 @@
...
+  - immutable content:  `/ipfs/{cid}`
+  - mutable content
+      - cryptographic, IPNS record signed with a specific libp2p key: `/ipns/{libp2p-key-cid}` 
+      - human-readable, [DNSLink](https://dnslink.io): `/ipns/{dns-name}`
+        (which in turn, resolves to either an immutable or mutable content path from DNS TXT record)
+
+
+**There are known gaps in the mutable story:**
+- IPNS records are only published by the original author, which severely limits the utility of our mutable names
+    - It is not possible to pin/republish `/ipns/{libp2p-key-cid}` names by other peers, so when original publisher disappears, IPNS links no longer resolve and we see "IPNS rot".
```
Stebalien, line 32: This isn't _entirely_ true, from a protocol standpoint, at least. If the IPNS author sets an infinite "EOL" on an IPNS record, anyone can republish. However, IPNS records expire by-default to limit how "stale" of a record an attacker could return to a client.

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,258 @@
...
+
+**There are known gaps in the mutable story:**
+- IPNS records are only published by the original author, which severely limits the utility of our mutable names
+    - It is not possible to pin/republish `/ipns/{libp2p-key-cid}` names by other peers, so when original publisher disappears, IPNS links no longer resolve and we see "IPNS rot". 
+    - **Due to IPNS rot people simply don't use it:**
+        - Instead publishing updates via `ipfs name publish {cid}`, website operators choose to constantly update DNSLink's DNS TXT record
+            - They point at reliable immutable paths (`dnslink=/ipfs/{cid}`) (no IPNS record means it resolves fast and never break).
+            - It works around the need for running a service responsible for constant republishing of IPNS record.
+            - This is a pragmatic choice, however:
+              - adds admin related to DNS updates (need to learn DNS provider's API for automating this, which introduces soft vendor-lock-in)
```
Stebalien, line 38: There are actually several "well known" APIs for updating "dynamic" DNS records. We should support those.

proposals/42-human-readable-mutable-names.md
```diff
@@ -0,0 +1,258 @@
...
+- **OS-level DNS resolver is used for all DNS TXT record lookups.** 
+  - IPFS node is unable to control which DNS resolver is used for specific TLD, all queries go to global DNS server configured in the operating system
+    - This harms competition and slows down innovation in the decentralized naming space. For example, if user changes their DNS resolver to one that is capable of resolving all TLDs over ENS, they are unable to resolve TLDs from OpenNIC or UnstoppableDomains. 
+  - Given how DNS is implemented in OS and ISPs, queries are most likely sent in plain text and can be spoofed in LAN or MITMd at the ISP level. At the time of writing this proposal DNS over HTTPS is not supported, and it is not trivial for regular user to set up a custom DNS server to change this behavior.
+
+**TLDR:** **We need to fix and future-proof the way mutable names work in our stack:**
+- Make IPNS useful on its own, and inside of DNSLink records.
+    - DNS TXT record should be set only once, updates should be published over IPNS.
```
Stebalien, line 47: Why? Purely so we don't have to have a way to auto-update DNS records?

Stebalien (commented):
Also see some prior work in https://github.com/ipfs/go-ipfs/pull/6560 (which unfortunately died in the bike-shed).

TL;DR: I'd really like to have this.

proposals/42-human-readable-mutable-names.md
lidel: While protocol is mostly ok, IPNS rot is a mixture of implementation choices:

- The default is not an infinite EOL, so everything expires unless original publisher bumps the record over and over again
- Expired records are ignored during resolution, even if no valid one is present and expired ones are all we have
- Only original publisher can bump EOL and keep record available by republishing it

proposals/42-human-readable-mutable-names.md
lidel: Mind providing examples, so I know we talk about the same thing? 

I think "several" is the problem here. I am not aware of any vendor-agnostic  API for updating DNS records that is supported across multiple DNS vendors. Everything is a soup of vendor-specific HTTP APIs and I don't think IPFS project should endorse vendor lock-in like  that.

proposals/42-human-readable-mutable-names.md
lidel: Correct. It removes the need for keeping DNS credentials and in context of ENS it removes the cost of paying for gas during each update.

proposals/42-human-readable-mutable-names.md
Stebalien: Hm. Yeah, I'm seeing the same thing. For simple A records, it looks like you can basically just "template" a URL. But that's not going to work for custom records.

There is RFC 2136, but it doesn't look well supported. I guess IPNS is simpler...

Gozala:
> @Gozala the ability to override OS-level resolver is crucial for both user choice/security (OS-level one is most likely plaintext from ISP) and for our browser integration story. Brave wants to add suport for ENS via DoH at https://eth.link/dns-query, with more systems to come, and I believe go-ipfs should provide primitives to follow DNS resolver choices user made via their user agent.
> 
> Being able to support non-ICANN names is something we get as a side effect of this work. We already support ENS in a hacky, hardcoded way in go-ipfs. What we propose to do here is to remove vendor-specific code and use DNS for interop.
> 
> My plan is to ship go-ipfs with some predefined DoH endpoints, so TLDs like `.eth` work out of the box, but the user has the choice to change the default resolver (right now, user has none).
> 
> As for DIDs, I don't believe there are any primitives for human-readable names, so out of scope here(?)

Thanks for elaborating, basically I would like if this was something that user could configure at OS level (even if in user space) so that brave, ipfs and others could recognize and respect as opposed to each coming up with own solution and creating a very complex user experience.

Gozala:
> @Gozala this is way beyond this proposal, but in the future go-ipfs could expose own DoH endpoint on localhost, so other services can use it as DNS cache/resolver. No need for additional service.

Maybe I'm missing something here, but what I was trying to imply is not sticking more things into ipfs config, but rather looking at things in e.g. `~/.config` that way other programs like brave could converge to doing the same. If other programs will need to have go-ipfs to do DNS resolution it's very unlikely to happen.

The reason I'm bringing this up is because there is a tendency to build things into ipfs, e.g. we have key manager, despite every OS having a native one. I realize that is not the case here, but talking to others teams might enable us to have a greater impact.

lidel:
This is a valid point, but feels out of scope: of this project

- splitting go-ipfs config, or at least extracting secrets like keys to separate files is unrelated implementation detail already tracked in https://github.com/ipfs/go-ipfs/issues/6262
- afaik no generic format for storing DNS resolution rules per TLD/wildcard exists, 
  and I don't see any immediate need for IPFS project to invent one:
  - Things like this take a lot of spec work and are more complex than they sound: even simple  `~/.config`  is not as simple as it look (macos, windows will have own conventions, and even linux expects  `XDG_CONFIG_HOME` support  that may override default location etc)
  - Browser vendors like Brave won't be interested in additional complexity that comes with supporting a new standard like this.    
    - User agents like Firefox or Brave want to be "the source of truth": 
      - Firefox has own CA store + one can set custom DoH resolver via `network.trr*` in `about:config` that overrides OS-level settings
      - In near future user will be able to set DNS resolver for ENS in Brave, which needs to have technical means of ensuring go-ipfs complies with that choice without bothering the user.
    - What we want to do in this PR is to make go-ipfs flexible enough to support Brave use case (dnslink in `ipns://`)  and enable experimentation with non-ICANN naming systems. 

**TLDR:** Introducing a OS-level config for DNS resolution is out of scope here, but could be added in the future, as a follow-up project ("if go-ipfs config does not have custom resolver defined, fallback to user/OS-level config").

lidel:
Implementation can be tracked in "Ongoing work" of https://github.com/ipfs/go-ipfs/issues/6532

PR #43, opened 2021-02-17
Create Distributed signaling.md

Gozala:
Is worked on at https://hackmd.io/@gozala/distributed-signaling

### DIFF
--- a/proposals/Distributed signaling.md
+++ b/proposals/Distributed signaling.md
@@ -1,0 +1,116 @@
+# Distributed signaling
+
+Authors:
+- @gozala 
+- @hugomrdias
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
+A proposal should contain enough detail for others to understand how this project contributes to our team’s mission of product-market fit
+for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on,
+and any other information relevant for prioritizing this project against others.
+It does not need to describe the work in much detail. Most technical design and planning would take place after a proposal is adopted.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). 
+Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+Centralized nature of signaling servers harms the ecosystem because:
+
+1. We actively tell everyone to configure nodes to talk to their servers.
+2. Teams building products have added burden of infrastrcuture.
+3. Network gets fragmented peers configured to use signaling serevers A aren't gonig to discover peers that are configured to use signaling servers B.
+
+
+What we want instead is ecosystem in which:
+
+1. New product teams can dedicate additional resources (as in operate always-on dialable nodes) without fragmenting a network.
+2. New product improving infrastructure of the network as a whole, as opposed to creating an isolated island.
+3. Remove a need for configuring nodes.
+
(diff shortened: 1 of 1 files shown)


raulk:
@Gozala I think what you're thinking of here is already covered in #21 -- Project Flare, much of which has already been implemented.

Gozala:
> @Gozala I think what you're thinking of here is already covered in #21 -- Project Flare, much of which has already been implemented.

I don't believe that is the case, e.g. WebRTC requires signalling server and NAT hole punching isn't going to address that. Today you need to configure node with an address of a signaling server, which is a problem because ethier:

1. We configure nodes so they all dial same address which
  1. Introduces central point of failure
  2. Challengers our ability to handle all the traffic it will create.
2. We continue telling users to configure their nodes to use their own infrastructure which:
  1. Adds complexity when building products
  2. Fragments network around signalling hosts

Ideally we could grow signaling infrastructure and make it possible for nodes to discover new signaling nodes as they join the network.

@raulk Am I misunderstanding Project Flare, or does my comment above makes the difference more clear ?

Gozala:
I did closer review of  Project Flare #21 and I think while there is a bit of overlap there it does not directly address the problems this proposal is aimed at. More specifically in WebRTC creating a `RTCPeerConnection` requires URLs for `iceServers` at the moment (unless I'm mistaken) it is pulled from the node configuration. Which implies that we either centralize & increase load on specific servers or we fragment the network.

Desire here is to change that such that

1. Nodes do not need to be configured (With list of ICE candidates)
2. Allow nodes to discover available ICE candidates
3. Allow other teams building on products to bring new ICE candidates to the network (as opposed to configuring own nodes to use separate infrastructure).

I focused on WebRTC here, but that is also the case with bootstrap nodes and more

raulk:
@Gozala I see. It wasn't clear this proposal is WebRTC focused.

> 1. Nodes do not need to be configured (With list of ICE candidates)
AFAIK, nodes don't need to be configured with ICE candidates, they discover their own candidates with the assistance of ICE servers, right?

> 2. Allow nodes to discover available ICE candidates
Did you mean to say "ICE servers"? ICE already handles candidate discovery...

> 3. Allow other teams building on products to bring new ICE candidates to the network (as opposed to configuring own nodes to use separate infrastructure).

Same. You mean "ICE servers"?

---

Basically this proposal aims to create a decentralised fabric of ICE servers for WebRTC nodes to rely on?

Gozala:
> Basically this proposal aims to create a decentralised fabric of ICE servers for WebRTC nodes to rely on?

No that was just an example. I think it is broader than that. E.g. we also configure nodes with addresses for boostrap nodes, if those nodes were to become unreachable by part (or worth whole) of the network they would not be able to recover without reconfiguring nodes, release of new version of products etc... Here solution would be to make nodes remember dialable nodes so they could be used when boostrap nodes can’t be dialed.

Broadly speaking we tell our collabs to configure nodes with addresses to their servers. I think that is harmful for the ecosystem and I’d like to figure out a way that would allow us to change our messaging from “configure your nodes to talk to your servers” to “spin additional nodes to ensure that network is reliable for your users”.

This draft is clearly needs more work to clarify all that.

vasco-santos:
@Gozala correct me if I am wrong, as far as I understand the goal here would be to assign roles to nodes and advertise these roles in the network. For instance, you would have a set of libp2p nodes which would be "ICE servers". These nodes would advertise (via DHT, Rendezvous, ..) that they have such a role and peers in the network can find these advertisements.

Considering the above, I think that this could be part of the Limited Relay work described in Project Flare. Ideally any node acting as a limited relay should be able to enable two peers with webRTC to exchange SDP offers, instead of having a special implementation of a relay to do this. As we are going to implement a new version of the circuit protocol, it would be good to also have in mind WebRTC signalling.

> No that was just an example. I think it is broader than that. E.g. we also configure nodes with addresses for boostrap nodes, if those nodes were to become unreachable by part (or worth whole) of the network they would not be able to recover without reconfiguring nodes, release of new version of products etc... Here solution would be to make nodes remember dialable nodes so they could be used when boostrap nodes can’t be dialed.

This is not the case anymore since we have a persisted PeerStore. We will try to connect to the configured bootstrap nodes + all previously known nodes (within the connectionManager thresholds).

What we really need to do here (at least in JS) is creating an intelligent way of doing follow up dials on restart:
- If we already know a lot of peers, why connect back to the boostrap nodes?
- If peer X is (in theory) not valuable to us why connect to it?
  - for example, if a given node has pubsub enabled, it would benefit from connection to peers that also run pubsub instead of the DHT

https://github.com/libp2p/js-libp2p/issues/744 check the proactive dial part in this issue

Gozala:
> @Gozala correct me if I am wrong, as far as I understand the goal here would be to assign roles to nodes and advertise these roles in the network. For instance, you would have a set of libp2p nodes which would be "ICE servers". These nodes would advertise (via DHT, Rendezvous, ..) that they have such a role and peers in the network can find these advertisements.

I think this is a great way to put it, thanks! I also think there will be roles beyond "ICE", e.g me and @hugomrdias talked about "IPNS record keepers" as one such a role.

PR #44, opened 2021-02-17
Proposal: Infrastrucuter coop

Gozala:


olizilla:
I'm very into this idea. Last time around, a proposal with a simpilar spirit to this ont got stuck on "ipfs.io" should stay the canoninical gateway to simplify recognition and tooling, but we didn't properly unpack what that means when there is also dweb.link and subdomain gateways.

The pervious converstation is here 
- https://github.com/protocol/bifrost/issues/10 
- https://github.com/protocol/bifrost/issues/12 
- and particularly the video linked to in https://github.com/protocol/bifrost/issues/12#issuecomment-521872459

it's on a private repo, so shout if you need access.

PR #45, opened 2021-02-17
Proposal: Unified FileSystem API

Gozala:
Worked on at https://hackmd.io/@gozala/unified-fs-api

### DIFF
--- a/proposals/Unified FileSystem API.md
+++ b/proposals/Unified FileSystem API.md
@@ -1,0 +1,116 @@
+# Unified FileSystem API
+
+Authors:
+- @gozala
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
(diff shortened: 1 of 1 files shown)


proposals/Unified FileSystem API.md
```diff
@@ -0,0 +1,116 @@
...
+When multiple apllications share an IPFS node they end up with shared local filesystem view, which is impractical because they can introduce race condition and access each others secret files.
+
+
+###### Content Froking
+
+Neither of the existing file system APIs are adequate for forking file tree and performing edits on it. Global filesystem API does not really has write API, and with local filesystem API file tree needs to be assigns a path, which not only requires adding file tree and removing afterwards but is also prone to race conditions.
+
+---
+
+What we need instead is an API similar to `ipfs.files` that can be instantiated with an arbitrary `CID`. This would address all the above use cases.
```
lidel, line 39: Hm.. we can do `ipfs files cp /ipfs/{cid} /sub-mfs` for unixfs DAGs and then mutate ` /sub-mfs`.
Perhaps what is missing is ability to copy non-MFS dags to  MFS, and porcelain for mutating them?

Gozala: I don't think copying into mfs is a good way to go about it. All the use cases I had was about building up a tree and then either adding to network or MFS (or both) in one operation. Adding things to mfs introduces room for race conditions and an overhead which is unnecessary, and it also requires cleaning up after.

PR #46, opened 2021-02-18
Proposal: store-my-text example app

rvagg:
@ZenGround0 gifted this idea, but I really like it as an example app because of the simplicity of the frontend which gets rid some of the complexity of a typical example app (no need for frontend frameworks!). The UI is primarily email-driven, which also lends itself to complexities of background deal-making for small resources.

The path from this example app to a genuinely useful application is pretty clear too - if we nail the basics, it could turn into a full NFT-storage application without much effort.

### DIFF
--- a/proposals/46-store-my-text-example-app.md
+++ b/proposals/46-store-my-text-example-app.md
@@ -1,0 +1,111 @@
+# Store-my-text Filecoin example app
+
+Authors: @zenground0 @rvagg
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/46
+
+## Purpose &amp; impact 
+
+#### Background &amp; intent
+
+_Describe the desired state of the world after this project? Why does that matter?_
+
+We have an extremely simple example application, _and_ a functioning service that has real-world utility that exercises Filecoin and deal-making.
+
+ * A **very** simple web application, minimal UI fluff and distraction: **Save my text to Filecoin**
+   - Text field to enter some UTF-8 text
+   - Field to submit your email address
+   - A button to "Submit"
+ * User enters some non-empty text and their email address and presses "Submit"
+ * Application performs a simple confirmation-of-email ("click this link") to confirm a real user and manage trivial DoS
+ * Application handles storage of text on Filecoin and sends confirmation email
+   - Application mediates the storage deal-making process for small content - finding miners, batching content, etc.
+ * CID provides persistent "proof-of-text" to the user
+
+Initially this is just a very simple system that demonstrates (and exercises, and educates us on) the full deal flow for simple content.
+
+There's plenty of extensions that could be made to this, turning it into a genuinely useful application, such as:
+
+ * IPFS integration
+ * Other input types (images, etc. getting toward the NFT use-case)
+ * Payments ("Deposit X FIL to this address...", "Enter your CC to purchase X FIL")
+ * ETH integration to turn content into an NFT (ERC721, ERC1155, whatever is needed, maybe integration with an existing NFT service?)
+
+#### Assumptions &amp; hypotheses
+
+_What must be true for this project to matter?_
+
+ * Deal-making is difficult
+ * Building applications that perform simple & small storage deals to Filecoin is currently difficult
+ * Web3 developers want solid but grokkable example applications they can copy (fork-and-go!)
+
+#### User workflow example
+
+_How would a developer or user use this new capability?_
+
+ * Play with application live (we host and manage and pay for storage of small amounts of text, with limits per email address)
+ * Fork open source application
+ * Make new web3 app with Filecoin storage
+ * Profit
+
+#### Impact
+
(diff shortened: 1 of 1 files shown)


proposals/46-store-my-text-example-app.md
```diff
@@ -0,0 +1,111 @@
...
+ * IPFS integration
+ * Other input types (images, etc. getting toward the NFT use-case)
+ * Payments ("Deposit X FIL to this address...", "Enter your CC to purchase X FIL")
+ * ETH integration to turn content into an NFT (ERC721, ERC1155, whatever is needed, maybe integration with an existing NFT service?)
+
+#### Assumptions &amp; hypotheses
+
+_What must be true for this project to matter?_
+
+ * Deal-making is difficult
```
vasco-santos, line 38:
```suggestion
 * when attempting to do deal, frequently it is difficult to get it completed
```

rvagg:
I need to add a disclaimer in here that the actual path to storing data in Filecoin isn't necessarily to take that all on directly, but it could involve third-party tools, like Powergate and even ThreadDB. I think my preferred approach is to aim for direct so we can get into the problem space and do some learning, but maybe we end up bailing because of all of the complexities of miner-selection, deal-making, etc. But it would be good to explore it as a learning experience I think.

yusefnapora:
This sounds like a great example 👍

> The UI is primarily email-driven

I could see "email to filecoin" becoming a useful product in its own right someday - the async nature of the deal flow is a really good fit for email status updates, and it could be a useful way to permanently archive conversations by forwarding them.

mgoelzer:
@rvagg 
> because of all of the complexities of miner-selection, deal-making, etc

I run a PL miner that you can use if all else fails, and I think I can help advise on deal-making as I have a lot of experience now going through the deal->store->retrieve workflow.  Keep me looped in and I'll keep some notes on what things were hard to figure out or get working; that'll be good learning for the project in addition to good learning for the team.

proposals/46-store-my-text-example-app.md
```diff
@@ -0,0 +1,111 @@
...
+ * Make new web3 app with Filecoin storage
+ * Profit
+
+#### Impact
+
+_How would this directly contribute to web3 dev stack product-market fit?_
+
+ * Exercise deal-making for small resources
+ * Provide a very simple example for web3 app developers to build on
+
```
mgoelzer, line 57: 1.  Team learning is legitimate and necessary; it contributes to PMF because we have to understand the developer journey (esp for new devs) before we can improve it.

2.  Product (me) can learn and use those findings to better identify the top problems facing new developers.

proposals/46-store-my-text-example-app.md
```diff
@@ -0,0 +1,111 @@
...
+   - Application mediates the storage deal-making process for small content - finding miners, batching content, etc.
+ * CID provides persistent "proof-of-text" to the user
+
+Initially this is just a very simple system that demonstrates (and exercises, and educates us on) the full deal flow for simple content.
+
+There's plenty of extensions that could be made to this, turning it into a genuinely useful application, such as:
+
+ * IPFS integration
+ * Other input types (images, etc. getting toward the NFT use-case)
+ * Payments ("Deposit X FIL to this address...", "Enter your CC to purchase X FIL")
```
mgoelzer, line 31: This one is important to do (does not have to be in rev 0).  It's going to reveal a whole other set of blockchain pain points that developers will face, and we should face them too, to see how we can improve.

dineshshenoy:
I like the flow!  One tactical consideration is the acceptance criteria.

I would imagine that the initial version would be built on testnet/nerpa net/caliabration net which has different characteristics than mainnet.

This [issue](https://github.com/filecoin-project/slingshot/issues/440#event-4327559089) talks about some of the hiccups storing small data sizes on mainnet with sector sizes of 32+ GB.

rvagg:
@dineshshenoy yeah, deal size is one of the big things that concern me about this, but I'm pretty keen for us to create an app that hits that hurdle so we can understand the shape of it. I'm fine with batching data prior to deal making - and the async user process lends itself to that - batching seems like a necessary feature of an app that builds on Filecoin and we have the IPLD skills to solve that elegantly. But if we have to batch simple text input until we get into the Gb range then, yeah, that's going to be a problem. Perhaps it simply means putting a pause on this project when we hit a show-stopper that we can't solve in reasonable time (and with reasonable _example_ code) and move elsewhere while the underlying problems are addressed.
And to be clear - retrieval isn't part of this initial scope, that's recognised as a separate thing to tackle but would be a great extension to this to make a complete system.

dineshshenoy:
Agreed - worth pursuing regardless.  Just wanted you to be aware of the roadblock (and try to hit it) since you may not hit it on nerpa net.

mikeal:
The ICE info is incomplete in the proposal. I doubt we’ll be able to 👍👎 without it.

PR #47, opened 2021-02-18
Proposal: JS libraries are trim and modular (build a plan of attack)

rvagg:
This builds on various people's recently expressed thoughts, including @achingbrain, @vasco-santos, @mgoelzer, @jacobheun, @mikeal and others. I've seen this same sentiment expressed in different ways but @achingbrain wrote it up fairly nicely yesterday and I've captured some of that here. I'd love for any of you who have additional thoughts on this to hack on my branch here to enhance this proposal if you think it's lacking.

Currently this is a proposal to build a plan of attack, not to do the work. I'm conscious of the need to keep the scope manageable so this is just initial work and would lead to further proposals to _do_ the work but likely in pieces, ranked by impact and dependence.

### DIFF
--- a/proposals/47-js-trim-and-modular.md
+++ b/proposals/47-js-trim-and-modular.md
@@ -1,0 +1,124 @@
+# JavaScript libraries are trim and modular
+
+Authors: @rvagg @achingbrain @vasco-santos
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/46
+
+## Purpose &amp; impact 
+
+#### Background &amp; intent
+
+_Describe the desired state of the world after this project? Why does that matter?_
+
+Using IPFS with JavaScript requires installing and loading a _lot_ of code and you end up with a monolith. Using js-ipfs necessitates it being _the primary thing_ in your application stack because of its size and complexity.
+
+1. There are a _lot_ of dependencies
+   1. Takes a long time to install
+   2. Assumes the user wants _all the things_
+2. The total weight of code is very large
+   1. Related to the _all the things_ accumulative design of the stack
+   2. Crypto libraries mostly to blame
+3. The API surface area is very large
+   1. Related to the projection of the CLI API up into the browser API and exposing _all the things_
+   2. The complexity of the entire stack means there's a lot of different ways to achieve things using different APIs and configuring them separately, but you still get _all the things_
+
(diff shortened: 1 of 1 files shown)


proposals/47-js-trim-and-modular.md
```diff
@@ -0,0 +1,124 @@
...
+
+Reach JavaScript developers at their level of comfort. 
+
+Surveys like [the one conducted annually by StackOverflow](https://insights.stackoverflow.com/survey/2019/#most-popular-technologies), suggest that there are nearly an order of magnitude more developers who know Javascript (67%) than who know Go (8%). But our JavaScript offering is not friendly for general consumption and we usually end up pushing them to use go-ipfs (and friends) and build wrappers around it. Many use-cases are much simpler than requiring the entire js-ipfs stack as it currently exists and we're not serving those (well).
+
+#### Leverage
+
+_How much would nailing this project improve our knowledge and ability to execute future projects?_
+
+#### Confidence
```
lidel, line 62: As a data point to increase confidence here, Textile did not use our libs, but built their own ipld+libp2p variant.

@Gozala was looking into creating something like this a while ago and ended up transfering the project from Textile to our org. 
It now lives at https://github.com/ipfs/js-dag-service

achingbrain: There was some great work there, this proposal would definitely take those learnings on board.

mikeal:
Looks great. We should also take into account some of the proposed work for Gateway and Pin APIs since they open opportunities for building small and pretty dumb clients that do a lot.

Gozala:
I really like this idea! I would love if it would could make other filesystem abstractions like ones implemented by fission.codes, textile teams as first class option for anyone to pick up as opposed to making them a second tier.

rvagg:
@carsonfarmer raised this today in the IPLD meeting and suggested that the DagService is a key part of composability for the work they're doing (although composability across the stack would be nice too!). Being able to extract that piece and/or swap it out for alternatives would be really helpful. There's some prioritisation work that can be achieved here that lets us chew off pieces of it that have impact while not committing to the whole lot.

proposals/47-js-trim-and-modular.md
```diff
@@ -0,0 +1,124 @@
+# JavaScript libraries are trim and modular
+
+Authors: @rvagg @achingbrain @vasco-santos
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/46
```
olizilla, line 5:
```suggestion
Initial PR: https://github.com/protocol/web3-dev-team/pull/47
```

PR #48, opened 2021-02-18
Proposal: Lotus can use external wallets

rvagg:
Mostly based on work by @dineshshenoy and feedback from @eshon regarding ecosystem partners and advice we need to give to web3 app developers. It seems like a small but high value feature to add that would supercharge our app developer ecosystem and help partners quite a bit.

I'm flying a little blind here, having no knowledge of the underlying mechanics involved other than the block layout of the messages (I can tell you all about block layouts, that's just not very helpful). So I'd really appreciate help from more knowledgable folks to get this right.

@magik6k I know you had some feedback on the version of this in the original Projects Map doc, maybe you have some easy input? It's a little different here, based on @dineshshenoy's version rather than using callbacks.

One idea is that team Sudo could tackle this as a way to get our heads deeper into Filecoin. We're a JS-heavy team so it's not going to be as simple as it might be for a Go team, but that might be part of the value-prop here as we could be getting our heads and skills deeper into the stack.

Alternatively this might be trivial for someone else to pull off!

### DIFF
--- a/proposals/48-lotus-external-wallet.md
+++ b/proposals/48-lotus-external-wallet.md
@@ -1,0 +1,189 @@
+# Lotus can interact with external wallets
+
+Authors: @rvagg @dineshshenoy 
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/48
+
+## Purpose &amp; impact 
+
+#### Background &amp; intent
+
+_Describe the desired state of the world after this project? Why does that matter?_
+
+Lotus has a number of operations that require messages to be signed before landing on chain. The current set of JSON-RPC calls as well as the Lotus CLIs do a number of tasks which simplify the process for the user:
+
+ 1. Assemble the JSON message
+ 2. Generate a nonce
+ 3. Sign the message
+
+These steps require internal state:
+
+ 1. A `From` address.
+ 2. A nonce (a monotonically increasing integer that’s associated with a particular actor)
+ 3. A key for message signing
+
+Storing these items internally make it difficult for scaling in certain circumstances as a wallet is associated directly with a Lotus node. Decoupling would activate many use-cases:
+
+ * Multiple wallets used by a single node to sign messages - shared node among multiple users
+ * Wallets fully detached from nodes, allowing for lightweight interaction with the chain by message signers
+ * Removing custodial requirements for nodes enables better security models, particularly for complex configurations (both miners and clients)
+
+It is atypical for blockchain nodes to have custody of keys on behalf of Dapp developers but we are currently forcing this, or forcing alternative architectures (such as node operators transferring FIL to their own wallets to perform transactions).
+
+We currently don't support the typical MetaMask + Infura style Dapp model and our partners and potential partners are asking for this decoupling.
+
+**Example:**
+
+Raw message generated from user input:
+
+```json
+{
+    "Version": 0,
+    "To": "f12qy5dbvss5wjpiucebkkp4hjixcmotnmzpx2bka",
+    "Value": "60000000000000000000",
+    "GasLimit": 611585,
+    "GasFeeCap": "101067",
+    "GasPremium": "100013",
+    "Method": 0,
+    "Params": null
+}
+```
+
+Currently lotus performs the following steps:
+
+* Add the `From` attribute
+* Add a `Nonce` value
+* Add a message CID calculated from the resulting assembled message (TODO: is this a CID of the original message? it can't be of the "assembled message" as that's self-referential)
+* Sign the message
+* Add a payload CID calculated from the resulting payload
+
+Resulting in the final message to be submitted to the chain:
+
+```json
+{
+  "Message": {
+    "Version": 0,
+    "To": "f12qy5dbvss5wjpiucebkkp4hjixcmotnmzpx2bka",
+    "From": "f1hbls2zmkfk7tzjre3rlv3faxh7n4ntb47dnii5i",
+    "Nonce": 6,
+    "Value": "60000000000000000000",
+    "GasLimit": 611585,
+    "GasFeeCap": "101067",
+    "GasPremium": "100013",
+    "Method": 0,
+    "Params": null,
+    "CID": {
+      "/": "bafy2bzacedu4nx3lzxf2njupasdqg5xl2imbyoa6ksib6cjyykgxyybb6nxbs"
+    }
+  },
+  "Signature": {
+    "Type": 1,
+    "Data": "X6EBxXARBRfMlOd89lFnC98RFlkV6ZYcaupLiIfopfVe32hpcgvQvNgp3+QA2+DtsBjK14S/ShiEFnvPrGpTTwA="
+  },
+  "CID": {
+    "/": "bafy2bzacebifjagyqbypbxjiekkclerj4apxwyucao2glmvi4ufm6olvw3wc2"
+  }
+}
+```
+
+There are two critical steps in deal-making that require signing:
+
+* Client signs a message to move funds into StorageMarketActor (to chain)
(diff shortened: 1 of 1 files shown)


Kubuxu:
I think a lot of people are calling feature like this: "Bring your own signature".

willscott:
Doesn't the [glif wallet](https://wallet.glif.io/) already do this? is this about making a library / re-exportable version of that? having an existing implementation should make this hopefully straightforward 🤞

eshon:
Storage Deals currently require Filecoin Lotus Node key signing.

Many blockchain developers are used to Ethereum’s Infura + MetaMask dapp model, where Infura provides hosted Ethereum node endpoints that anyone can use. Dapps are designed so that Metamask or other Web3 wallets holding the private keys client-side can be used to sign transactions that are sent to generic Infura or other trusted endpoints to the Ethereum network.

Currently in Filecoin that model does not work for storage deals. Token transfer transactions can be signed by external wallets via the Lotus API + MPoolPush(), but storage deals require a full Filecoin node’s keys.

Currently, we send dapp developers to Powergate, which custodies FIL on behalf of users on the server and maps multi-tenancy user wallets to a Powergate instance mapped to the Filecoin Lotus Node's key. However crypto asset custody is not ideal for any service provider.

rvagg:
Yes, so there do seem to be two separate things in here which _could_ be separated but the solution to both could end up being basically the same (if they are both to be "solved").

 1. A general "Bring your own signature" for message signing, any messages - currently it's possible to author and sign messages directly and submit them via `MPoolPush`, which I suppose is what glif and others are doing right now. But we could potentially make that process easier such that the user doesn't need to do all of the message crafting.
 2. The deal funding + proposing + making dance requires a series of messages along with state and Lotus currently doesn't give you any good way to interact with that or handle parts of it manually (from what I understand).

Solving for **1** alone seems low value, good client libraries can solve this I think.

Solving for **2** is high value, and depending on the solution it may also provide additional options for **1** along the way.

The two solutions I've seen are (as I understand them):

 1. Inserting a callback mechanism in the deal flow.
 2. Enabling a "no-publish" mode in the deal message creation flow that lets the caller take the messages, sign them and submit them signed (directly via `MPoolPush` or perhaps there's a nicer API interaction there).

Kubuxu:
I think we have a different view for "Bring your own signature", for me it is about Lotus producing message prototypes in places where it would just send out the message today.
This allows you to "bright your own signature" to the party by signing messages Lotus is proposing.

eshon:
>Solving for 1 alone seems low value, good client libraries can solve this I think.

Yes, signing Tx's external to a node is already possible now using Zondax's [Filecoin Signing Tools](https://github.com/Zondax/filecoin-signing-tools) Library which Glif uses, it has multisig and paych too

rvagg:
@Kubuxu:

> I think we have a different view for "Bring your own signature", for me it is about Lotus producing message prototypes in places where it would just send out the message today.

I think that's what I'm trying to say in my second proposed solution. But I'm speaking from a lot of naivety here and I'm sure my language and terminology is imperfect. When a team (or person) takes on this task, it'll probably need input from others about the best way to implement it with lotus, and the most idiomatic way for our ecosystem--to @eshon's point.

mikeal (approved):

raulk:
@Kubuxu could you elaborate on how much of this is enabled by the new message prototype APIs + inversion of control that we introduced in the Message Management project? My understanding is that we'd want a mode where Lotus APIs that send a message just return the raw payload for the caller to sign and push directly to the mpool.

raulk:
Caught up with @Kubuxu today.

- The plumbing for external signatures is in place in https://github.com/filecoin-project/lotus/pull/5822 🚀 
- Merging the above is blocked on the v1 API: https://github.com/filecoin-project/lotus/pull/5685
- Some methods that used to send messages now return a message prototype, concretely the multisig methods. This is the basis for signing externally.
- Methods that send messages but still haven't been adapted include: markets, deal making, payment channels.
    - The reason is that these domains all have some form of a "manager" abstraction that complicates the refactor a little.
- For external signing to be useful in the context of Bedrock and Nitro, we need at least markets and deal making JSON-RPCs to be adapted.

The way to perform external signing is:
1. Client calls a method that returns a message prototype.
2. Client obtains a nonce through JSON-RPC. This is not automatic because the API doesn't know if the client is batching messages or sending a one off.
3. Client obtains a gas estimate.
4. Client fills in the nonce and gas estimate in the message prototype, and creates a message payload.
5. Client signs the message payload with a self-provided signature.
6. Client invokes MpoolPush to push the signed message.

We believe that the delta between the status quo and seamless external message signing would be as follows (in order of priority):

1. To create an example go app demoing external signing, that can serve as a reference for other implementors / users. Right now we can only do this for raw messages (e.g. transfer) and multisig messages.
2. Migrate markets JSON-RPCs to message prototypes.
3. Migrate deal making JSON-RPCs to message prototypes.
4. Migrate payment channel JSON-RPCs to message prototypes.

eshon:
Not sure this is the best place to mention this but:

With respect to deal making, external signature deal flow should ideally rely on a proper escrow so it can be trustless. If the deal falls through, the clients should be able to get their funds back.

@ribasushi said this doesn't happen with the current StorageMarketActor: funds are not returned to clients if the deal aborts pre-publish. So perhaps in a later v2 an escrow Actor could be considered.

ribasushi:
> said this doesn't happen with the current StorageMarketActor: funds are not returned to clients if the deal aborts pre-publish. 

This is my understanding fwiw, that can be very well be very wrong. From what I've been told there is a "gray area" in the storage deal flow where funds get forfeit by both parties. If correct in the first place, it wouldn't be a local/remote signature issue, but rather "SMA flow" issue.

Fatman13:
Hello, @raulk,

For...

> The way to perform external signing is:
> 
> - Client calls a method that returns a message prototype.
> - Client obtains a nonce through JSON-RPC. This is not automatic because the API doesn't know if the client is batching messages or sending a one off.
> - Client obtains a gas estimate.
> - Client fills in the nonce and gas estimate in the message prototype, and creates a message payload.
> - Client signs the message payload with a self-provided signature.
> - Client invokes MpoolPush to push the signed message.

`venus-messager` has a working production implementation of this work flow. A slight difference is that `venus-messager` handles all the nonce assignment, gas estimation and message publishing within the `messager` module itself and there is no much back and forth with the client.

> To create an example go app demoing external signing implementing the flow above. It can serve as a reference for other implementors / users. Right now we can only do this for raw messages (e.g. transfer) and multisig messages.

Such demo app already exists in form of `filwallet`, where all messages can be signed by an offline wallet (phone).

Fatman13:
Hello, @raulk,

For...

> The way to perform external signing is:
> 
> - Client calls a method that returns a message prototype.
> - Client obtains a nonce through JSON-RPC. This is not automatic because the API doesn't know if the client is batching messages or sending a one off.
> - Client obtains a gas estimate.
> - Client fills in the nonce and gas estimate in the message prototype, and creates a message payload.
> - Client signs the message payload with a self-provided signature.
> - Client invokes MpoolPush to push the signed message.

`venus-messager` has a working production implementation of this work flow. A slight difference is that `venus-messager` handles all the nonce assignment, gas estimation and message publishing within the `messager` module itself and there is no much back and forth with the client.

> To create an example go app demoing external signing implementing the flow above. It can serve as a reference for other implementors / users. Right now we can only do this for raw messages (e.g. transfer) and multisig messages.

Such demo app already exists in form of `filwallet`, where all messages can be signed by an offline wallet (phone).

BigLep:
Nitro will reevaluate priority in Q3.

PR #50, opened 2021-02-18
Proposal: Service mocks

rvagg:
For fast unit & integration testing against our stack.

### DIFF
--- a/proposals/50-service-mocks.md
+++ b/proposals/50-service-mocks.md
@@ -1,6 +1,6 @@
 # Mock implementations for our key services
 
-Authors: @rvagg
+Authors: @rvagg, @vasco-santos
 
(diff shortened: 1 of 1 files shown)


willscott:
this'll be easier to evaluate with a concrete list of the mocks to create, in terms of estimating how long it'll take

rvagg:
@willscott yep, the scope is probably too large here and the implementable version of this involves just picking one, maybe a simple one, or a narrow use-case of one, to start with. Perhaps there's a specific flow with the Lotus API that requires a small subset of the API and a limited set of behaviours that could be packaged up into a useful library. Requires investigative work and feedback about user stories / needs I think.

mikeal:
we should put this in a “Draft” state until the list of mocks is flushed out.

PR #52, opened 2021-02-19
Proposal: Free retrieval via IPFS

alanshaw:


hannahhoward:
@alanshaw

a couple of thoughts --
1. Bitswap is running on the LOTUS node already and used for chainsync. Right now only one blockstore can be written to with Bitswap per libp2p host. I believe the host in question here though would be a lotus-storage-miner process which isn't using bitswap so I think it works?
2. Currently by default the miner keeps a sealed piece, and copy of the unsealed piece, and now they would be adding a third copy to live in a blockstore IPFS can talk to. Seems like a lot to ask? I guess they can price accordingly. But one downside is other than the sealed piece, I have no way to verify you're living up to your side of the bargain -- i.e. you're not proving to me you haven't deleted the unsealed piece, and/or cleared out your blockstore. So it'd be a bummer to pay a premium for retrievability on IPFS with no guarantee it actually would be retrievable that way.
3. I think in general I favor the other approach of making IPFS speak Filecoin, i.e. as Myel is proposing, as opposed to this direction of having Filecoin speak IPFS... though I will admit I lack a clear justification for why I feel this way.

alanshaw:
Thank you for taking the time to read and comment!

>     1. Bitswap is running on the LOTUS node already and used for chainsync. Right now only one blockstore can be written to with Bitswap per libp2p host. I believe the host in question here though would be a lotus-storage-miner process which isn't using bitswap so I think it works?

😅 we'll find out I guess! Another option might be graphsync - I _think_ I remember this being added to IPFS already...

>     2. Currently by default the miner keeps a sealed piece, and copy of the unsealed piece, and now they would be adding a third copy to live in a blockstore IPFS can talk to. Seems like a lot to ask?

I found out about fast retrieval (i.e. the copy of the unsealed piece) after writing this. Do you know if it would be possible to use that store?

> I guess they can price accordingly. But one downside is other than the sealed piece, I have no way to verify you're living up to your side of the bargain -- i.e. you're not proving to me you haven't deleted the unsealed piece, and/or cleared out your blockstore. So it'd be a bummer to pay a premium for retrievability on IPFS with no guarantee it actually would be retrievable that way.

Yes - valid concern. Do we have _any_ retrievaibility checks right now?

>     3. I think in general I favor the other approach of making IPFS speak Filecoin, i.e. as Myel is proposing, as opposed to this direction of having Filecoin speak IPFS... though I will admit I lack a clear justification for why I feel this way.

🙏 In my experience gut feelings are a good tell. It would be good to nail down some justifications though. I think there is value in prototyping this out, if only to help with nailing them down.

proposals/52-free-retrieval-via-ipfs.md
```diff
@@ -0,0 +1,157 @@
...
+
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+* PoC is accepted and a robust solution is added to Lotus and released.
+* 1 or more collabs run Lotus with this option enabled.
```
aschmahmann, line 118: This seems a bit rough since there is 0 discovery in this proposal. This means that the only way for a standard go-ipfs node to retrieve this data is if it knows the exact multiaddr of the peer. It might be worthwhile to join the public IPFS DHT simply for peer routing so that at least go-ipfs nodes only have to know the peerID of the target and not the full multiaddr.

proposals/52-free-retrieval-via-ipfs.md
```diff
@@ -0,0 +1,157 @@
...
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+* Enabling bitswap on a Lotus node might increase memory/CPU usage to the extent that it would affect it's normal operations.
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+Attacking this from the other angle and upgrading IPFS with abilities to store and retrieve from Filecoin. e.g. [myel.network](https://myel.network/).
```
lidel, line 128:
```suggestion
Attacking this from the other angle and upgrading IPFS with abilities to store and retrieve from Filecoin. e.g. [myel.network](https://myel.network/), [proposal #57](https://github.com/protocol/web3-dev-team/pull/57).
```

proposals/52-free-retrieval-via-ipfs.md
lidel: Well, there are pretty good options for discovery:
- pinning service backed by Lotus could return multiaddrs in  `delegates` field from [pinning services api](https://ipfs.github.io/pinning-services-api-spec/)
- pinning service could also suggest setting up [Peering](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#peering) on nodes that require Filecoin retrieval (IPFS nodes used as hot cache would be peered with Filecoin nodes)

lidel (commented):
I like this because:

- Bitswap interop would enable [Peering](https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#peering) setup where IPFS node acts as hot cache and Filecoin nodes provide retrieval of archived data. 
- Pinning services could return Lotus multiaddrs in `delegates` field ([api spec](https://ipfs.github.io/pinning-services-api-spec/))
- Filecoin talks IPFS, not the other way around. 
  Does not open  Pandora's box of adding Filecoin to IPFS :grimacing:

mikeal (approved):

PR #57, opened 2021-02-24
Initial ipfs/filecoin retrieval proposal

hannahhoward:
[Rendered](https://github.com/protocol/web3-dev-team/blob/feat/ipfs-filecoin-integration/proposals/filecoin-retrieval-in-ipfs.md)


proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+
+2. The IPFS Client sees a record for paid Filecoin retrieval in the DHT and uses the paid Filecoin retrieval plugin to talk to the Lotus Lite HTTP api to generate parameters for a paid retrieval deal. The IPFS client then initiates a data transfer request. As needed, the plugin calls out to Lotus Lite HTTP API to generate additional vouchers.
+
+Future example: Custom retrieval payment system for a "retrieval miner" using a custom payment system
+
+1. I negotiate out of band with a custom paid retrieval provider an agreement to retrieve data.
+
+2. I use the IPFS Pinning API to request to pin data. The miner stores the data in Filecoin as backup, and to prove it's being stored, but then distributes several copies to it's CDN pinning network and advertises all records in the IPFS DHT as paid retrievals that use its custom protocol.
+
+3. To retrieve, I use the information in the DHT and a plugin supplied by the provider to generate vouchers for a data transfer with one of their miners close to me.
```
aschmahmann, line 66: Generally 👎 for plugins. They don't work on Windows and we have lots of users that use Windows. They can be useful for prototyping, and are fine whenever users are happy to build the binaries themselves, but otherwise not particularly good.

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+- A DHT change discussion is a big design discussion, and could take a long time or hit roadblocks. It also could involve breaking changes to the content routing API in Libp2p, which might cause a lot of headaches
+- It may turn out to be quite difficult to make smart decisions about when to use Filecoin retrieval vs regular Bitswap. The simple case where data is pinned ONLY in Filecoin and not IPFS is easy to decide on. When its in both, how we decide what to use is not totally clear.
+- The long term future of Filecoin retrieval protocol in general is dependent on improvements to proofs that enable faster unsealing or reliable systems for gauranteeing unsealed copies. (i.e. Filecoin Plus)
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+See [Proposal: Free retrieval via IPFS ](https://github.com/protocol/web3-dev-team/pull/52). If all of that proposal's assumptions turn out to be true and it is possible to just have miners throw data in a bitswap accessible blockstore, it is probably the fastest path to baseline filecoin retrieval in IPFS.
```
aschmahmann, line 128: If we want free retrieval without running a local Filecoin client then IMO that proposal seems like the thing to do.

However, the more complete solution to this problem requires a local Filecoin client and a proposal closer to this one

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+See [Proposal: Free retrieval via IPFS ](https://github.com/protocol/web3-dev-team/pull/52). If all of that proposal's assumptions turn out to be true and it is possible to just have miners throw data in a bitswap accessible blockstore, it is probably the fastest path to baseline filecoin retrieval in IPFS. 
+
+There are as mentioned various variations within this proposal -- building a seperate DHT for Filecoin retrievals as opposed to extending the IPFS DHT, having a data transfer voucher HTTP API Spec as opposed to a local plugin system, etc.
+
+#### Dependencies/prerequisites
+
+This project should commence after the currently in-progress [IPLD in IPFS project](https://github.com/protocol/web3-dev-team/pull/25) as they are hard to do simultaneously.
+
```
aschmahmann, line 135: Depending on how many provider records would need to be emitted per-miner we might need to do some of the optimization on DHT providing first.

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+#### What does done look like?
+_What specific deliverables should completed to consider this project done?_
+Done is when you can complete a roundtrip pinning / retrieval to Filecoin from IPFS and a spec exists for building future paid retrievals
+
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+- Well known pinning services begin offering Filecoin based pinning
```
aschmahmann, line 112: This isn't strictly required as you can pin directly to a Filecoin light client, although it's a nice way to see if things are catching on.

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+- Accelerator projects around new kinds of retrieval begin building plugins for paid retrieval
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+- A DHT change discussion is a big design discussion, and could take a long time or hit roadblocks. It also could involve breaking changes to the content routing API in Libp2p, which might cause a lot of headaches
```
aschmahmann, line 121: Yes, depending on the implementation this may require some work on two fronts:
1. Deciding how we want to upgrade/bump the DHT protocol version. We have some ideas here, but AFAIK we've never done it. This is something worth doing anyway.
2. Making + testing the DHT changes

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+High
+
+We know multiple companies that wish to use Filecoin as the storage provider for IPFS data. We know that many have been unsuccessful so far in bridging the gap.
+
+## Project definition
+#### Brief plan of attack
+
+1. We work with libp2p & IPFS stewards to come up with design for how we put filecoin retrieval records in the DHT. Note: I have used language so far that assumes these records will go in the IPFS DHT itself, but it may make sense to explore putting them in a seperate DHT that IPFS also has access too.
+
+2. We work out a way in Lotus to get these DHT records posted as part of the deal making process.
```
aschmahmann, line 94: Note: records have to be continuously reposted too, and some combination of the miner and client need to decide what records are worth posting.

aschmahmann (commented):
Overall I like the idea of having a Discovery and Fetching API in go-ipfs that allows getting data from other systems like Filecoin.

This proposal makes the APIs pretty specific to the state of Filecoin today, but I don't know that it'd be that much more work to make it more general.

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+
+Example: IPFS pinning with Filecoin, using remote pinning API and free retrieval
+
+1. As an IPFS user who wants to pin data with filecoin, I negotiate an agreement to pin data in Filecoin with a pinning service out of band in exchange for an access token (i.e. I pay them through their website for X GB of filecoin pinning at Y Price). 
+
+2. I use the remote pinning API from the IPFS command line to pin some data with them.
+
+3. The pinning service initiates and completes a filecoin deal to store the data
+
+4. The pinning service, upon completion posts a record in the IPFS DHT for the CID of pinned data that contains both the peer address for the miner and also flag to indicate it requires filecoin retrieval (and potentially also includes the PieceCID?).
```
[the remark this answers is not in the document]
willscott: The benefit of having additional schemes in DHT is that we don't have an explosion where we search every possible provider for each piece of content - first we learn which protocol to use, and then we make use of it.

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+
+5. For retrieval, upon querying the DHT, the IPFS client sees the the record is a filecoin retrieval record, and instead of using standard bitswap fetching, switches over to proposing a free retrieval via the data transfer protocol to the miner, and completes the retrieval this way.
+
+Future Example: Paid retrieval with IPFS & Lotus Lite client on same computer
+
+1. I make a deal with a miner through my lotus lite client to store some data. When my deal is done, the miner OR client can advertise the peer address and retrieval parameters in the IPFS dht 
+
+2. The IPFS Client sees a record for paid Filecoin retrieval in the DHT and uses the paid Filecoin retrieval plugin to talk to the Lotus Lite HTTP api to generate parameters for a paid retrieval deal. The IPFS client then initiates a data transfer request. As needed, the plugin calls out to Lotus Lite HTTP API to generate additional vouchers.
```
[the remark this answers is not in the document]
willscott: we should figure out what the plugin (or api) interface looks like. authenticated API over which cid+selector is requested seems compatible with this.

proposals/filecoin-retrieval-in-ipfs.md
willscott: I'm less worried about plugin vs rest vs ipc as long as there's a clear interface :)

proposals/filecoin-retrieval-in-ipfs.md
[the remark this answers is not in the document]
hannahhoward: I think there may be a misconception on what the data transfer module is here.

Data transfer is ideally an ABSTRACT protocol for organizing data transfers. The voucher system is a mechanism for defining authentication and payment schemes. It is not tied to filecoin retrieval - in fact we use it already for filecoin storage. The transport layer, while tied to graphsync currently, does not need to be. There is already in Data Transfer an abstract notion of a tranport interface. Currently the only protocol with an implementation to satisfy it is go-graphsync, but the intent is that we could support other protocols.

The idea here is to bring in data transfer, and initially satisfy graphsync + filecoin retrieval, but to potentially support any payment scheme on any transport.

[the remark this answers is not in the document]
hannahhoward: we may also honestly need to augment data transfer to accept alternate targets other than cid+selector.

proposals/filecoin-retrieval-in-ipfs.md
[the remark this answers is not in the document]
aschmahmann: I'm assuming then you say "every possible provider" you mean every possible discovery system. If so then sure, it's nice if there can exist a single global routing system with all the data so you don't have to worry about what order to check systems in or whether to waste resources checking them in parallel. In this scenario we need option 1, more metadata in provider records.

However, it seems likely that multiple routing systems are going to emerge and that one size will not fit all so we'll have to worry about the multiple routing systems problems anyway. Making the routing systems more easily extensible (option 2) will allow for more experimentation in this space.

hannahhoward:
@aschmahmann this may not be as specific to Filecoin retrieval as it appears. I'd love to go over the data transfer module with you some time -- both what it does today and what it could do in the future. It is ideally supposed to be a general purpose module for ANY kind of data transfer.

Whether using plugins, IPC, or HTTP APIs, the intent is for external provides to be able to supply any kind of retrieval, with essentially built in case being free filecoin retrieval.

lidel (commented):
Hm.. I know it is probably intended to be a generic solution, but current version does not read like that at all.
It reads like adding Filecoin-specific protocols to IPFS, and that feels really wrong to me: goes against the underlying philosophy behind IPFS, libp2p, IPLD, Multiformats being distinct, reusable open projects without any vendor lock-in. 

My gut feeling is that unless we have more than one use case (Filecoin retrieval), we should not add anything that complex to reference implementation of IPFS.  If Filecoin is built on top of IPFS, then Filecoin should implement generic IPFS protocols, not the other way around. 

Added pre-mortem inline.

proposals/filecoin-retrieval-in-ipfs.md
pooja: Where is the spec for what Filecoin light clients can do? My hope with light clients is that they are purely lightweight interfaces to the Filecoin network and don't do any direct storage/pinning themselves

aschmahmann: There isn't one in particular that I know of. #40 is sort of there but is likely heavier than what you'd expect.

I wasn't being very clear here, so I'll try and clarify. My thought was that some reasonably light Filecoin client could be asked to store data on Filecoin using the remote pinning API, it would download the data from IPFS and then send it off to the miner who is ultimately storing the data. At that point the Filecoin client could just delete the data. If the Filecoin client + IPFS nodes are on the same machine this seems like it could be pretty efficient.

proposals/filecoin-retrieval-in-ipfs.md
```diff
@@ -0,0 +1,163 @@
+# Filecoin Retrieval In IPFS
+
+Authors: @hannahhoward
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+Problems Addressed:
+
+- Data that has been stored successfully is not guaranteed to be retrievable. (P1)
```
dchoi27, line 9: In terms of how the proposal solves this problem, is the assumption that lotus nodes being able to add DHT records and IPFS to retrieve from lotus nodes will improve retrieval guarantees because there will be more demand for retrieval (so more reason for lotus nodes to supply the data)?

proposals/filecoin-retrieval-in-ipfs.md
yiannisbot: Is it reasonable to make the assumption that retrieval miners are more stable than IPFS nodes and therefore records don't need to be republished as frequently as they currently do in IPFS?

aschmahmann: The republishing interval is a DHT-wide parameter derived from the network churn. Unless there's a separate DHT that mostly contains stable nodes, or there are enough retrieval miners to influence the properties of the public IPFS DHT it doesn't help our republishing intervals

mikeal (approved):

hannahhoward:
attn: @willscott @pooja @lidel @aschmahmann 

This proposal aligns with Golden Path/Bedrock, but it was posted a while ago and is really out of date. I've gone back and made a significant effort to update it so it matches with current plans and knowledge gained over the last few months.

The key changes are:

1. I've removed the content about using IPFS remote pinning APIs and narrowed the scope to cover only the process of retrieving data. The proposal was too big and the parts about using remote pinning were very unclear.

2. Per @lidel and @aschmahmann , I've made a serious effort to make this proposal as generic as possible in terms of the software that goes in IPFS. There are no adds to IPFS other than go-data-transfer. I believe we can support free retrievals without adding any Filecoin specificity, and we can define a generic RPC interface difference providers can offer to support arbitrary paid retrieval protocols. I believe we can do this without making the Filecoin bridge, which I know is our biggest product priority, any more difficult to implement.

3. The strategy for free retrievals is to put the initial data transfer voucher directly into the DHT, which then means there is no additional Filecoin specific software in IPFS other than data transfer needed to make a free retrieval.

Also, I believe there are two components of this project that themselves entire projects:

1. Either upgrading the IPFS DHT to support these records or implementing a composable routing interface to support fetching these records from a centralized index and interleaving them with regular DHT results so from the point of view of higher level software its all one stream of records.

2. Refactoring go-ipfs and go-bitswap so that you can make a DHT query, then decide between bitswap vs data transfer for fetching.

I am going to write up a proposal for second of these. I suggest @aschmahmann or @petar should be in charge of first.

BigLep:
Just for bookkeeping before the proposal for "Refactoring go-ipfs and go-bitswap so that you can make a DHT query, then decide between bitswap vs data transfer for fetching" gets written up, there are some great notes and diagrams in https://hackmd.io/JoZiAAtnTpqAKuQaEUra4g

mikeal:
We’re having discussions between several teams on the DHT side of this, working with ResNetLab (@petar) to get requirements flushed out and with Gateway (@olizilla ) to get some of this running in production sooner rather than later since the Indexer work doesn’t touch content discovery until the last mile of that work stream. Let’s make sure all of these efforts across teams are aligned.

PR #59, opened 2021-03-02
worker work-pool

coryschwartz:


proposals/multi-miner-worker-pool.md
```diff
@@ -0,0 +1,109 @@
...
+
+####  What does success look like?
+Worker nodes are 100% utilized even with uneven miner allocation.
+
+#### Counterpoints &amp; pre-mortem
+I'm not confident that this is the biggest problem big miners have.
+It will be more complex to maintain two different kinds of job scheduling.
+
+#### Alternatives
+* Workers could be made to connect to multiple miners simultaneously. Operators with multiple miners would try to schedule on workers with the least-load, or round-robin, hash-ring or some other mechanism and re-try when there is too much load on the worker.
```
magik6k, line 95: The storage/sealing logic is written in a way which should support accepting tasks from multiple miners, even on separate networks with different sector sizes

It should be as simple as making it possible to run the "storage manager" piece as a standalone process, and making it possible to tell lotus-miner to use an external storage manager  (or even multiple manager instances, giving us worker grouping for free)

PR #60, opened 2021-03-02
Proposal: Clearing house / batching service protocol

rvagg:
I'm sure others have thought along these lines before and have more mature ideas in their head than me so I'd love to hear feedback and know whether there's good reasons to _not_ pursue something like this. It's particularly aimed at serving the small-deal case but could serve deals of any size.

What I'm not sure about is channel-conflict with ecosystem partners who are pursuing similar things. I haven't used Powergate but I believe the Textile suite might be doing something roughly in this direction, but I'm under the impression that's intended for single-tenant workflows where the same user is batching their own data and you have to buy into their whole suite to make it work nicely.

### DIFF
--- a/proposals/60-clearing-house-batching-service.md
+++ b/proposals/60-clearing-house-batching-service.md
@@ -110,6 +110,13 @@ _How might this project’s intent be realized in other ways (other than this pr
 
 #### Future opportunities
 
+**Protocol integration with Filecoin**
+
+Iteration of this protocol toward a trust-less state, where the user of such a service has guarantees of funds-paid to services-rendered could be achieved by deeper integration with Filecoin. Such as:
+
+* Adding a Filecoin primitive (actor) that can ascertain whether a Piece has been included in one or more sectors with some minimum expiration epoch.
+* Client payments to the broker could be handled via a Filecoin payment channel that releases payment only when a specific piece of data is present in some sectors (using the primitive above).
+
 ## Required resources
 
 #### Effort estimate
--- a/proposals/60-clearing-house-batching-service.md
+++ b/proposals/60-clearing-house-batching-service.md
@@ -1,0 +1,128 @@
+# Filecoin Clearing House / Batching Service protocol
+
+Authors: @rvagg
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/60
+
+## Purpose &amp; impact 
+
+#### Background &amp; intent
+
+_Describe the desired state of the world after this project? Why does that matter?_
+
+* Small deals are hard, and economically unattractive for miners; anecdotally, with current costs, attractive deals are in the order of 16 GiB for standard and 4 GiB for FIL+
+* Many use cases involve parties with _only_ small deals to make and little opportunity to batch their own data
+* Dealing with fr32 padding combined with needing to have deals be power-of-two sized will generally lead to waste on the client side (e.g. [see this table for optimal deal sizes](https://github.com/filecoin-project/filecoin-docs/issues/705#issuecomment-787618483))
+* The deal-making latency is quite large, in the order of 12-24 hours for many deals to be completed and sealed
+* Deals require relationships with miners, the overhead of acquiring and maintaining relationships is a negative for developers building on Filecoin
+* The overhead of managing redundancy, monitoring deal-lifecycle, renewing deals, etc. is a negative for developers building on Filecoin
+
+Batching deals to achieve optimal deal size for miners should enable a better matching of incentives between client and miner. Batching can either be done at the application layer, or it could be done as a service, aggregated across many applications. A batching service **"protocol"** could define a standard API and workflow that could be adopted by layer 1 ecosystem partners and present a common interface to application developers and a simpler set of concerns and rules required to store assets on Filecoin.
+
+This proposal is for developing a protocol and an operational reference implementation.
+
+* The batching service is responsible for making the deals direct with miners. Deals would be large in size, ideally 16 GiB or 32 GiB and would be FIL+.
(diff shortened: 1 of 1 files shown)


proposals/60-clearing-house-batching-service.md
```diff
@@ -0,0 +1,128 @@
...
+
+This proposal is for developing a protocol and an operational reference implementation.
+
+* The batching service is responsible for making the deals direct with miners. Deals would be large in size, ideally 16 GiB or 32 GiB and would be FIL+.
+* Users of the service only need to establish a relationship with the batching service, not with miners (for storage, retrieval is out of scope for this initial proposal).
+* Users make **micro-deals** via the batching service, depositing FIL with the service in batches to be consumed in small amounts for each deal.
+* Users submit data via an API or IPFS (or other means) and the data is placed in a queue.
+* The batching service aggregates enough data to fit into an optimal deal size, pre-determined for the service, perhaps with a compromise where it will make smaller deals when it hasn't received enough data in a certain number of hours.
```
jnthnvctr, line 28: So one complication here (which may vary from service provider to service provider) is that the economics are still sort of weird: 
1) The time guarantee for sealing could leave a sector mostly empty unless the provider has a guaranteed set of deals coming through (or like an average flow rate)
2) The "cost" for paying the diff between the cost of sealing for a miner + the fil collected may be higher for small files. 

I think there's a few arguments that could be made - either that cost is borne by the service provider (provided their making sufficient profit), or it could be borne by the miner (cost of maintaining a relationship with the provider).

proposals/60-clearing-house-batching-service.md
```diff
@@ -0,0 +1,128 @@
...
+
+* The web3 stack (Filecoin in particular) would be well served by a well defined protocol for batching services that are able to aggregate data and make deals on behalf of end users/developers
+* That definition of such a protocol and reference implementation(s) is not going to cause channel-conflict with existing ecosystem partners who are working in this area (e.g. Powergate) but may enhance their offerings by standardising an interface and workflow for batching
+
+#### User workflow example
+
+_How would a developer or user use this new capability?_
+
+* Create relationship with batching service - signup using DID of some form - get an API key
+* Deposit FIL with the service to be consumed over micro-deals made
```
jnthnvctr, line 60: How are you thinking about the FIL pricing here?

proposals/60-clearing-house-batching-service.md
```diff
@@ -0,0 +1,128 @@
...
+####  What does success look like?
+
+_Success means impact. How will we know we did the right thing?_
+
+#### Counterpoints &amp; pre-mortem
+
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+* This places an intermediate layer between miner and client which introduces potential safety and reputational concerns. A miner can't do a plain filter by CID in this mechanism since the actual data CID will be buried inside the deal (unless Lotus looks for filtered CIDs deep inside the DAG?). A batching service instance would have to have a relationship with its miners and would offload some of the filtering requirements. So miners needing baseline guarantees about content would need to ensure the batching service they deal with imposes that on clients.
+* There may be concerns with adjacent content inside single deals if your content is places next to objectionable content. I'm not sure how this plays out, but CID+selector where you manually adjusting the selector slightly gives you someone else's content might have unfortunate side-effects?
```
jnthnvctr, line 101: This might be a good q for the Murmuration folks, but I believe for things like Bitscreen they're basically adding content to a do not retrieve list -> which maybe @ribasushi  has some recollection if this could be easily altered to support CID+selector?

jnthnvctr, line 101: I think for the MVP you're probably ok, with some basic information re this. 

You could imagine a different offering that would offer a more expensive store only your stuff option (UberX) vs. the batching service (UberPool)

ribasushi: Um... selectors apply during retrieval alone. I think @rvagg was more concerned "my content is in the same PieceCID as this other stuff" being more problematic than "my deal for content is in a sector with other deals for that other stuff"

mikeal:
This would integrate quite nicely with nft.storage https://github.com/protocol/web3-dev-team/pull/62

anorth:
This (and #62) are awesome, thanks. I've been thinking about something similar.

As a subsequent iteration, I think we could reduce the trust levels required by:
- adding a Filecoin primitive (actor method) that can ascertain whether a Piece has been included in one or more sectors with some minimum expiration epoch
- client making payment to the broker via a payment channel that releases payment only when a specific piece of data is present in some sectors (using the primitive above)

rvagg:
Fantastic ideas @anorth, deeper integration with Filecoin itself hadn't even occurred to me but I suppose it should have. I've tended to think of the state of the VM as fairly static but there's a lot of potential for opening up to additional primitives that could support all sorts of novel workflows outside of the core.

I've added text in "Future opportunities" to capture your ideas. (Also feel free to edit this branch directly if you have more to add, I don't mind).

jnthnvctr:
I started to put together some thoughts on from the Clients side what I think they might expect from this service [here](https://www.notion.so/Batching-Provider-Requirements-bcb50316e52b47d2a6fb321aa5b354e9) (gathering more feedback as well)!

mikeal:
we’re going to need more ICE info in this proposal to weigh it against other “small deals” proposals.

rvagg:
🤷 I'd like to lean on @jnthnvctr to help fill those out and guide this forward - if this _should_ go forward.
"Effort" estimation might require some tighter scoping. I also suspect that there's sub-projects in here that may be a better way to frame this project.

PR #61, opened 2021-03-02
Proposal: DealScript

rvagg:
Just the basic idea so far to get it out of my head.

A miner proposed something simpler but along these lines on Filecoin Slack and I think pinged @eshon about it, maybe for a grant. There's some demand for configurability among miners. And the opaqueness of the deal-making process really sucks right now.

### DIFF
--- a/proposals/61-deal-script.md
+++ b/proposals/61-deal-script.md
@@ -1,0 +1,104 @@
+# DealScript
+
+Authors: @rvagg
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/60
+
+## Purpose &amp; impact 
+
+#### Background &amp; intent
+
+_Describe the desired state of the world after this project? Why does that matter?_
+
+The current deal-making process is blunt. Miners get to set their public ask parameters, which include:
+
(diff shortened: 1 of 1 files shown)


rvagg:
This is also related to a thread @whyrusleeping started about padding small deals. It should be straightforward for a client to say "I'll pay you extra to store my valuable, but small thing" and a miner to be able to handle that. Currently there's no way to have such a conversation with a miner. There's also no sane way to pad out your thing to make it bigger and acceptable to a miner--other than throwing empty bytes on the end of it and then having to transfer those empty bytes via graphsync just to pack it out to size. Why can't I pay you to store my 1 Mb file in your entire 32 GiB sector at the price of that entire sector?
If we had a better mechanism for miners to charge large amounts for small files, then you wouldn't even need to do that. "Why yes I'll take your 1 Mb file, but it'll cost you just as much as if you were asking to store 32 G". A miner should even be able to just have a flat deal fee - remove the per GB part of the deal specifier. "All deals with me will cost `0.000000018` FIL per epoch, regardless of size".

PR #62, opened 2021-03-02
Proposal: nft.storage for NFTHack

alanshaw:
This is a proposal for time limited storage of NFT data on IPFS, backed by Filecoin and provided **free** to [NFTHack](https://nfthack.ethglobal.co/) participants during the hackathon.

[Rendered](https://github.com/protocol/web3-dev-team/blob/70484a4ac8bf998465d3962d88112004b4e3bfc6/proposals/62-nft-storage-for-nft-hack.md)


ribasushi (approved):
Looks great!
Probably a few devils will be in the details, but that's to be hashed out during design.

lidel (approved):
This proposal is fully packed!  Provides clear onboarding path for NFT developers  that:
- use CIDs identifiers for addressing assets
- leverage IPFS for hot storage and data retrieval 
- provides a demonstration how Filecoin-backed pinning service could look like
- promotes future-proof api with  no vendor-lock in: 
  - makes [pinning services api](https://ipfs.github.io/pinning-services-api-spec/) the standard for long term persistence 
  - developers can switch from `nft.storage` to Pinata or something else without changing anything in their code (!)
  
+  current feature gaps  like adding `pin.remote` commands to js-ipfs are solved as implementation detail of this proposal.

proposals/62-nft-storage-for-nft-hack.md
```diff
@@ -0,0 +1,184 @@
...
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+This is a proposal for time limited storage of NFT data on IPFS, backed by Filecoin and provided **free** to NFTHack participants during the hackathon (March 19 2021).
+
+* Register `nft.storage` (or other cool domain name) and expose a [pinning service API](https://blog.ipfs.io/2021-02-19-go-ipfs-0-8-0/#remote-pinning-services).
+
+    This will either be a proxy to [Pinata](https://pinata.cloud/) (or other pinning service) or use PL's own [pinbot](https://twitter.com/ipfspin).
+    
+    This gives us space to experiment. Data storage and retrieval is de-risked by the service we pin the data to. It affords us a playground with "_real_" data that we can store on Filecoin.
```
jnthnvctr, line 29: Super crucial! I think right now a lot of our collabs (e.g Foundation) are looking to us to have well designed solutions that they can drop into - and are looking to us to define how these things shoudl work. Being able to experiment here would be super valuable!

proposals/62-nft-storage-for-nft-hack.md
```diff
@@ -0,0 +1,184 @@
...
+
+    Here's some of the AWESOME things we could build and deploy in this space:
+
+    * Create a library that automatically makes deals on Filecoin for the pinned data (use/adapt the dealbot?). This is mentioned in the [AWS Facade proposal](https://github.com/protocol/web3-dev-team/pull/34) and could be re-used there.
+
+    * Implement a [deal batching service](https://github.com/protocol/web3-dev-team/pull/60) prototype for increased successful deal probability.
```
jnthnvctr, line 35: I think this is going to be super important as more folks like Textile/Fleek are enabling IPFS -> Filecoin services. Having a reference implementation would be powerful. 

One note - we should apply for DataCap with Andrew ( through https://plus.fil.org so we can be set up to make deals if we need to well in advance!)

proposals/62-nft-storage-for-nft-hack.md
```diff
@@ -0,0 +1,184 @@
...
+
+    * Run a "[Free Retrieval via IPFS](https://github.com/protocol/web3-dev-team/pull/52) Lotus node" that `nft.storage` can make deals with and that will expose the NFT data to IPFS.
```
jnthnvctr, line 37: Perhaps related to the next two bullet points, but should we consider also trying to get an MVP service for "CID/Deal querying"? 

I think one pattern that has emerged with NFTs is we have these end users (you or me) who use an app like foundation who will manage the storage of data as a blackbox-> but we run into the "if X was hit by a bus, do I lose my data". 

One answer to that ( which I think @rvagg  touched on for his batching proposal) - was potentially exposing the CID/Deal info so a user could grab that externally (and take over retrieval and payment) - is it worth trying to also prototype something lightweight here?

proposals/62-nft-storage-for-nft-hack.md
```diff
@@ -0,0 +1,184 @@
...
+This will cement IPFS as the primary means of off-chain NFT storage and will raise awareness of Filecoin as a storage provider.
+
+Currently a lot of NFT data is stored on IPFS with no clear story for permenance. Leveraging the remote pinning service API and providing guarantees of availability for the duration of the hack will prompt developers building NFT related software to think about permenance and who provides it. It will present Filecoin as an option for consideration and will likely drive some traffic to pinning services like Pinata.
+
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+* NFT users want to store NFT data on IPFS.
+* NFT users want guarantees of data availability.
```
jnthnvctr, line 71: I would say one others is that NFT users (though perhaps these should be called platforms to not overload a term) care about their end users (Artists) being able to take control of their data.

I think for an MVP we could define this as "An artist would be able to retrieve their data even if X app went away tomorrow"

proposals/62-nft-storage-for-nft-hack.md
```diff
@@ -0,0 +1,184 @@
...
+#### Leverage
+_How much would nailing this project improve our knowledge and ability to execute future projects?_
+
+<!--
+Explain the opportunity or leverage point for our subsequent velocity/impact (e.g. by speeding up development, enabling more contributors, etc)
+-->
+
+We may see an uptick in the number of developers working with NFTs that use IPFS for storage.
+
+The component part of actually storing the NFT data on Filecoin has invaluable potential to inform us on real world usage of the API's involved with making a deal. It should enable us to more confidently build solutions knowing they are the right thing to build.
```
jnthnvctr, line 101: +1000

mikeal:
This is a fantastic proposal 😍

proposals/62-nft-storage-for-nft-hack.md
alanshaw: > I think one pattern that has emerged with NFTs is we have these end users (you or me) who use an app like foundation who will manage the storage of data as a blackbox-> but we run into the "if X was hit by a bus, do I lose my data".

I'm not sure I understand what X is in this context. The storage provider? I would assume that the remote pinning service is doing it's best not to lose my data. Just like I assume Dropbox is doing their best to not lose my data. Perhaps this means multiple redundant copies are stored on Filecoin and I can see value in a library what negotiates deals to achieve this (RAID for Filecoin, except D - Disk is instead Deal 😉).

In IPFS this problem is alleviated a little - anyone that has ever accessed your CID will become a host for it (until they next GC) and can pin it themselves to keep the content alive on the network forever. So even if you only pin to a single provider that gets hit by a bus then there's a _chance_ you'll still be able to get your content from someone else in the network.

> One answer to that ( which I think @rvagg touched on for his batching proposal) - was potentially exposing the CID/Deal info so a user could grab that externally (and take over retrieval and payment) - is it worth trying to also prototype something lightweight here?

So, you'll have the CID of the content immediately. I don't think that it would be possible to take over the deal but you could negotiate a deal yourself with a different miner for the same content.

Assuming the content is _currently_ retrievable, there's nothing stopping you from pinning it yourself to a backup pinning service or your own IPFS node.

In the remote pinning service API you can query the status of a pinned CID. I think it would be a good idea to enhance that response to include details of the Filecoin deal it is currently backed up in so that it can still be retrieved even if the pinning service went away.

olizilla:
I love it! I would add the work we did on cube as a reference... having a UI where any and all members of the community can chose to contribute to the hosting cost for a given CID, and you can see how many years it has on clock, would be rad. The first n years could be covered automagically.

mikeal (approved):

proposals/62-nft-storage-for-nft-hack.md
```diff
@@ -0,0 +1,187 @@
...
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+This is a proposal for time limited storage of NFT data on IPFS, backed by Filecoin and provided **free** to NFTHack participants during the hackathon (March 19 2021).
+
+* Register `nft.storage` (or other cool domain name) and expose a [pinning service API](https://blog.ipfs.io/2021-02-19-go-ipfs-0-8-0/#remote-pinning-services).
+
+    This will either be a proxy to [Pinata](https://pinata.cloud/) (or other pinning service) or use PL's own [pinbot](https://twitter.com/ipfspin).
```
momack2, line 27: I think we should proxy to Pinata. I don't want the ecosystem to feel like we're creating a tool that eats someone else's business and reduces their ability to serve users in the ecosystem

PR #63, opened 2021-03-02
Create One-Click for Lotus on DigitalOcean Droplets

johndmulhausen:
Create one-click on DigitalOcean Marketplace based on AMI builds from https://github.com/filecoin-project/lotus/pull/5604 and document the workflow of using the image to get started.


coryschwartz:
By the way this is already merged.

https://github.com/filecoin-project/lotus/pull/5604


This PR creates an AMI and digitalocean snapshot.

The AMI works, however I think there's some additional work to do to get digitalocean snapshots to be available on the marketplace.

PR #64, opened 2021-03-02
proposal: IPFS PubSub based content routing

aschmahmann:
cc @Stebalien @petar


proposals/ipfs-pubsub-content-routing-v1.md
```diff
@@ -0,0 +1,156 @@
...
+
+<!--Explain why this rating-->
+3 . Among users with a large amount of data figuring out how to make that data findable is a top request.
+
+## Project definition
+#### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+The idea is that hubs will each subscribe to a pubsub channel (e.g. `/ipfs/content-routing/1.0.0`) and when users are looking for who has content they'll publish to that pubsub channel a request for the CID they are looking for. When a hub receives an inbound request for a CID they will check if they have it and if so they'll connect back to the user who will then end up fetching the data via Bitswap.
```
rvagg, line 80: If all of the items below were implemented, then we could also add another layer on top of generic content-routing for specific topic content routing, correct? So e.g. a video sharing service could sit on top of the general IPFS implementation and use these same mechanisms for finding CIDs but not crowd out the general space, `/ipfs/vidshare-contentrouting/1.0.0`.
So maybe this proposal is about the primitives for content routing, with a general implementation but also the possibility of focused topic use-cases too? A dapp could then just use all of these primitives for discovering CID providers and fetching content but be subscribed to an alternative broadcast channel.

aschmahmann: Yep, for efficiency we'd probably want to modify pubsub to be able to simultaneous broadcast messages across multiple topics without repeating itself. It's part of the spec and something we previously had pseudo-implemented, but if there's actual value there then it's probably worth implementing.

PR #66, opened 2021-03-03
propose: fast lane ipns

Gozala:


proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+
+Name owners could include addresses to chosen name keepers in the [DNSLink][] record, which would enable name resolution without resolving an address through name routers first.
+
+##### Visual illutrations
+
+###### Name resolution
+
+Diagram illustrates a name resolution flow. Note however that it does not show how name router can discover new names and their keepers from pubsub.
+
+<!--
```
Gozala, line 105: image

proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+![name resolution](./fast-lane-ipns/resolution.svg)
+
+
+
+###### Name update
+
+Diagram illustrating how name update flow.
+
+<--
+```flow
```
Gozala, line 149: image

proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+![publish](./fast-lane-ipns/publish.svg)
+
+
+###### Humane readable name resolution
+
+Diagram illustrates name resolution through DNSLink.
+
+<--
+
+```flow
```
Gozala, line 172: image

proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+>
+> - ❌ DHT in web nodes is impractical as most nodes will be undialable and so will be a web node.
+> - ❌ DHT implementation in nodejs is buggy, untersted and unoptimized ([ipfs/js-ipfs#3469][]). 
+> - ⚠️ PubSub in web nodes only works in swarms formed around central ICE server that are disjoint from go nodes. In practice creating fragmented networks of ephemeral peers.
+> - ⚠️ IPNS over pubsub is disabled in go by default (Can be enabled by running daemon with `--enable-namesys-pubsub`). Pubusb itself is also disabled by default ([protocol/web3-dev-team#53][]).
+> - ⚠️ IPNS over pubsub in nodejs still requires DHT to bootstrap overlay network of participating peers. Lack of proper DHT implementation prevents it. Which could be overcome via delegated routing but it does not work out of the box.
+
+Only DHT based implementation in go works out of the box. But it still fails to meet user expectations because:
+  1. *Slow* publish and resolution.
+  2. Requires continues republishing. (see [ipfs/go-ipfs#4435])
```
Stebalien, line 54: So does pubsub.

proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+3. Can become aware of name keeper changes through PubSub.
+
+> Note that since name keeper changes are expected to be rare TTL can be really high.
+
+Name routing service does not has to be limited to infrastructure nodes, peer could provide that service and resolve name keeper address from a local cache.
+
+
+###### DNSLink Optimization
+
+Name owners could include addresses to chosen name keepers in the [DNSLink][] record, which would enable name resolution without resolving an address through name routers first.
```
Stebalien, line 97: How would this interact with IPNS? IPNS doesn't use DNSLink.

proposals/fast-lane-ipns.md
Stebalien: I'm not sure where `a.com` is coming from.

proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+
+Teams that have tried using IPNS in their products either switched to using [DNSLink][] or opted-into IPNS over pubsub and still find it *(too slow)* for certain updates *(Textile)*.
+
+To meet a market fit for products with changing state requirement _(that is anything interactive)_ IPNS needs to work reliably, fast _(comparable to web2 solutions)_ and across all supported enviroments _(go, web, nodejs)_.
+
+##### Proposition
+
+At the high level this proposal promises an efficienecy of centralized publishing and resolution while retaining resilience of distributed system.
+
+Proposed name resolution system is inspired by Domain Name System (DNS) as it introduces athoritative name resolvers _(from now on referred as)_ **name keeper** nodes. But unlike DNS, **name owners** _(private key holders)_ are in charge of choosing a name keepers and retain ability to change a name keeper in the future.
```
Stebalien, line 64: How does this actually work? We could add a set of "keepers" to the IPNS record itself, I guess (as long as it's under a signature). I assume that's what you mean, right?

proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+
+> ❌ - Not available
+> ✅ - Available
+> ⚠️ - Not out of the box
+>
+> - ❌ DHT in web nodes is impractical as most nodes will be undialable and so will be a web node.
+> - ❌ DHT implementation in nodejs is buggy, untersted and unoptimized ([ipfs/js-ipfs#3469][]). 
+> - ⚠️ PubSub in web nodes only works in swarms formed around central ICE server that are disjoint from go nodes. In practice creating fragmented networks of ephemeral peers.
+> - ⚠️ IPNS over pubsub is disabled in go by default (Can be enabled by running daemon with `--enable-namesys-pubsub`). Pubusb itself is also disabled by default ([protocol/web3-dev-team#53][]).
+> - ⚠️ IPNS over pubsub in nodejs still requires DHT to bootstrap overlay network of participating peers. Lack of proper DHT implementation prevents it. Which could be overcome via delegated routing but it does not work out of the box.
```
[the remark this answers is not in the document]
Gozala: That sounds reasonable and possibly worth adding to considered alternatives section. Here I describe current state of things and to my knowledge it does so accurately, isn't it ?

proposals/fast-lane-ipns.md
Gozala: Conversation with @aschmahmann left me with an impression that pubsub is less problematic as peers query network and anyone can respond with a record, which in practice happens to be a lot more effective.

@Stebalien Anything I should change about above statements or was that just a side note ?

proposals/fast-lane-ipns.md
Gozala: I think I go into much greater detail below, but general idea is that instead of having to route on every name resolution, we'd do it very infrequently because there will be specific authoritative node you can ask about current record. That way nodes will be able to resolve "name keeper" address once and cache for future name resolutions.

Hypothesis is that while name keeper can change, it will be infrequent enough that we can see huge gains.

proposals/fast-lane-ipns.md
```diff
@@ -0,0 +1,336 @@
...
+###### Initial resolution optimization
+
+To make a very first name resolution nearly as efficient as subsequent ones it is propsoed to develop a new **name routing** service and deploy it on boostrap nodes. Name routing nodes will maintain `IPNS name → multiaddrs[]` mapping enabling peers to resolve an address of a name keeper on a first name resolution. Under assumbiton that reassigment of *name keepers* is going to be rare, overhead on name routing nodes should be insignificant as:
+
+1. Name routers could cache addresses for a long time.
+2. Resolving peers could cache address locally (and require router only first time).
+3. Name routers can deny service to peers resolving name too frequently.
+4. Name routers are not involved in publishing.
+
+It is assumed that name routers will be infrastructure nodes, implying that they can be dialed from limited nodes like browsers _*(Have SSL certificates setup, or have a WebRTC address)*_ and can leverage DHT and PubSub so thay they:
```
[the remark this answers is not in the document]
Gozala: I think rendezvous nodes could be a perfect fit here, I just did not wanted to tie the proposal to any specific not yet available (as in deployed) solution so it is possible to iterate on this without been blocked on other pieces.

proposals/fast-lane-ipns.md
Gozala: This is just DNSLink optimization, that would allow you to avoid querying routing nodes described in a previous section. When you resolve IPNS name you'd either already have name keeper address or you query routing nodes to obtain one and cache it forever.

proposals/fast-lane-ipns.md
Gozala: > I'm not sure where a.com is coming from.

As per test above this "Diagram illustrates name resolution through DNSLink", so it is an example when peer is trying to resolve `/ipns/a.com/`. @Stebalien does this make sense, otherwise could you please clarify a bit what you mean ?

proposals/fast-lane-ipns.md
Stebalien: IPNS and DNSLink aren't really related. For DNSLink, you just query your local DNS resolver which is almost always very fast.

proposals/fast-lane-ipns.md
Stebalien: https://github.com/protocol/web3-dev-team/pull/66#discussion_r599061000

proposals/fast-lane-ipns.md
Stebalien: It's less problematic as long as someone who resolved the record is still around and still subscribed to the record's pubsub channel. So, for popular records yes.

But for unpopular records, no, you'll need to be online and/or republish regardless.

proposals/fast-lane-ipns.md
Stebalien: Ah, sorry, my question is: how are you specifying these "name keepers"? Where does the list live, how is it published, how is it authenticated, how do I revoke keepers, etc.

Gozala: I think you may have meant how nome owners choose name keepers. I don't have thought through the specific here, but general idea is that name owner would provide a name keeper with a signed certificate, which it can use to prove to routing nodes that it is indeed the name keeping authority. 

That said I'm pretty open to discussing / reconsidering many details  of this proposal, mostly I'm pitching a bigger picture, that is lets make name publishing fast point to point operation. And make name resolution as close to that as possible.

proposals/fast-lane-ipns.md
Gozala: Oh I think I know what the misunderstand is here. Today most DNSLink records are `/ipfs/...` while expectation with this done is that whey will turn into `/ipns/...` and will not require DNS provider specific APIs to do updates.

Does this makes more sense ?

proposals/fast-lane-ipns.md
Stebalien: Got it. My thinking is that the keepers should be added as additional fields to the IPNS records. That way, there's no way to "roll back".

Gozala: > how are you specifying these "name keepers"?

1. You can create DNSLink record with `/ipns/${key}` and multiaddrs to name keeper.
2. You issue a certificate to a name keeper authorizing them as a name keeper. Which they use to let routing nodes know & proof they are the authority for the name. General assumption here is that pinning services will assume "name keeper" role as they are always on, well connected and you tell them what to pin anyway.

> Where does the list live

"Initial resolution optimization" attempts to describe it at high level. It assumes that infrastructure nodes can keep a list of "name keepers" and respond to queries from the network. If the name is unknown they can fallback to regular resolution DHT/PubSub and then cache the record forever. Note that here they resolve a "name keeper" not an actual record, and because those likely to change rarely no significant load is expected since there multiple layers of long term caching.

> how is it published, how is it authenticated

Name keepers are in charge of publishing to DHT, and over pubsub (possibly rendezvous) announcing themselves as "name keeper" for a name. Again since name keepers are unlikely to change TTLs could be very high.

> how do I revoke keepers, etc.

Again have not thought through all the details, but broadly speaking "name owner" would issue new certificate overriding old one to a different "name owner", which will than be responsible for updating the network. I expect pubsub would be an effective way to let routing routing nodes about the name keeper change. That said I'm open to alternative here as well.

Gozala: I realize I'm pretty vague when I talk about certificates, that is because I intentionally have not spend time thinking through all the details until this gets some traction. That said I think [UCAN](https://blog.fission.codes/auth-without-backend/)s are providing a pretty good and generalized model.

proposals/fast-lane-ipns.md
Stebalien: That's the first I've heard of that. What's the motivation?

Gozala: > That's the first I've heard of that. What's the motivation?

"What does success look like?" section calls out specifically related goals as success metric. That said motivation specifically is:

1. Browser nodes can not update name (unless....)
2. Not all DNS providers provide programatic API to do record updates. And ones that do have their own custom API so there is no general interface to publishing updates.
3. Usually updating DNS record involves separate authentication, creating yet another hurdle for doing it especially in browsers but also beyond.
4. We could have subscribe style interface for realtime stuff if domain name is just a human readable alias to `/ipns/...`, which seems unrealistic otherwise.

If I'm not mistaken originally https://github.com/protocol/web3-dev-team/pull/42 covered a lot of this before it rescoped.

mikeal (approved):

jacobheun:

PR #67, opened 2021-03-04
propose: wrap up rendezvous implementation and deployment

vasco-santos:


proposals/rendezvous.md
```diff
@@ -0,0 +1,156 @@
...
+#### User workflow example
+_How would a developer or user use this new capability?_
+<!--(short paragraph)-->
+
+Rendezvous can offer a wide range of scenarios.
+
+1. A developer will start a `js-ipfs` node in the browser (pubsub is enabled by default) and behind the scenes the node will be able to discover and connect to nodes that share the same interestes (pubsub subscriptions)
+2. A developer will start a `js-ipfs` node in the browser and the node will try to find available relays out of the box, so that it can use them to receive incoming dials.
+3. A developer can use the Rendezvous API to register peers under a given namespace, so that other nodes can easily find them. Example: Discover all the peers running in the dapp scope
+
```
Gozala, line 67: I wonder if slate use case is worth surfacing here.

vasco-santos: Slate would be included in the third use case I think.

Once dapps not rely on star servers, they need a way to discover peers on a dapp scope, so all the peers would register on a rendezvou server under a `slate` namespace and discover the peers that previously registered. 

Is there any other use case you have in mind?

Gozala: I meant actually mentioning the slate in the 3rd use case as a real use case makes it more obvious than hypothetical one

proposals/rendezvous.md
```diff
@@ -0,0 +1,156 @@
...
+_Success means impact. How will we know we did the right thing?_
+
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+- JS and GO nodes can out of the box establish meaningful connections to enable its subsystems to properly work
+  - Nodes are connected to 6 pubsub nodes
+  - Nodes are connected to X closest peers
+  - Browser nodes are always advertising a multiaddr through a relay
```
Gozala, line 117:
```suggestion
  - Browser nodes are always advertising a multiaddr through a relay
  - Browser nodes are dialed from go nodes
```

proposals/rendezvous.md
vasco-santos: Yes, that is a good call.

Gozala:
> I have some thoughts, but we should advance first with PL infrastructure and monitor performance before making these a community infrastructure. I did these benchmarks: [libp2p/js-libp2p-rendezvous#8](https://github.com/libp2p/js-libp2p-rendezvous/pull/8) (worth checking the last commit as this PR is done on top of the actual implementation)

I'm not sure we're talking about the same thing here. I am not asking if PL can subsidize community with rendezvous nodes, but rather interested in defining how community can add their rendezvous node without fragmenting the network (or the rendezvous registry).

vasco-santos:
Maybe I was not explicit enough. I was mentioning that deploying and orchestrating these rendezvous servers might not be easy, which would be problematic for "how community can add their rendezvous". These will likely have a large amount of requests and we need to understand if they will scale as expected before trying to grow them.

Regarding network fragmentation, we can think about a small protocol extension. We could have rendezvous servers registering on themselves (and others they know) as providing the "rendezvous server service". This would enable libp2p nodes to find all the rendezvous servers in the network. With this, libp2p nodes acting as rendezvous clients can start asking for peers on a given namespace to a subset of these rendezvous server until they have enough.

Gozala:
> Maybe I was not explicit enough. I was mentioning that deploying and orchestrating these rendezvous servers might not be easy, which would be problematic for "how community can add their rendezvous". These will likely have a large amount of requests and we need to understand if they will scale as expected before trying to grow them.

I would like to offer a slightly different perspective than one I think you might be picturing. There is on overlap between the role [SSB Pubs](https://ssbc.github.io/scuttlebutt-protocol-guide/#pubs) play in the network and rendezvous nodes in IPFS. I as far as I know large loads on pubs is not a problem in SSB, because they provide service to elected peers and peers also learn and connect to multiple pubs. This creates a nice, distributed system without much of an orchestration.

I would love if rendezvous was not designed under assumption of several nodes serving all of the network motivated by philanthropy (or worth yet, discovered a way to take advantage of their position), but rather many limited nodes serving the needs of specific groups or us cases motivated by specific cause.

mikeal (approved):

vasco-santos:
> This creates a nice, distributed system without much of an orchestration.

> I would love if rendezvous was not designed under assumption of several nodes serving all of the network motivated by philanthropy (or worth yet, discovered a way to take advantage of their position), but rather many limited nodes serving the needs of specific groups or us cases motivated by specific cause.

Yes, I totally agree this with be a really big improvement to the Rendezvous Spec. How should we add this to the proposal? I wonder it really depends on how fast we want to land Rendezvous. If we go with an alternative first, we can iterate on the spec and spend some more time on this

jacobheun:

PR #69, opened 2021-03-05
JS-driven storing from IPFS-API to Filecoin

willscott:
[Rendered](https://github.com/protocol/web3-dev-team/blob/471187a83406d92db62d610c6f958e04b8c31ce8/proposals/filecoin-pinning.md)

### DIFF
--- a/proposals/filecoin-pinning.md
+++ b/proposals/filecoin-pinning.md
@@ -1,0 +1,123 @@
+# [outcome or objective here] 
+
+Authors: @willscott
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
+A proposal should contain enough detail for others to understand how this project contributes to our team’s mission of product-market fit
+for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on,
+and any other information relevant for prioritizing this project against others.
+It does not need to describe the work in much detail. Most technical design and planning would take place after a proposal is adopted.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). 
+Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+Currently, to make sure data is persistently stored in a decentralized mechanism as an IPFS user is a complicated process.
+In this position, you need to get access to a filecoin node, either by running a full lotus node, or paying someone else, like
+infura to run one for you. Then, on the lotus command line, you can make a 'deal' with a miner to store data.
+The lotus deal process allows the source of the data to be a local IPFS node, and will then retreave and stream the data
+to the miner for storage, but there's no way to 'push' this data to the filecoin system from IPFS even though doing
+so is conceptually similar to the existing 'remote pinning' API that does already exist for centralized pinning providers.
+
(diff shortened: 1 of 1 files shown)


rvagg:
It seems to me that this is trying to point to the missing pieces of the chain that flow from dapp to Filecoin. Some of these have been covered in various ways on other proposals too, but in isolated ways, so maybe we're getting a better shared understanding of what's required.

I think the list of pieces includes:

* Proper IPFS<>Filecoin bridging, both ways
* Fully detached wallet interactions with Lotus and/or Lotus Lite, particularly to enable remote deals
* APIs to connect all of the pieces, including via JS/browser to enable full dapp style interop

There's probably some things about JS-IPFS browser<>server interactions that need to go into here too to complete the picture since we seem to have rough edges there. I know @achingbrain is working on a proposal to improve the IPFS HTTP API that might come in to play.

Deal size might be an issue too, which might bring batching-service layering into play as well.

Does that sound about right?

willscott:
Agree there are several pieces we should try decomposing this into.

* what does bridging IPFS<>Filecoin look like?
  * both use libp2p. can we get data to the miner via a direct bitswap/graphsync session from an ipfs client if we explicitly dial them and start a session? If the miner makes the request, does this look to the IPFS client any different from another ipfs node requesting the dag?
* In writing this, i _think_ the main opinion i'm positing is that the most interesting place to build a light library for deal making & wallet signing is in the browser with JS. I think this is lighter than lotus-light is currently imagined - and is closer to what a metamask-style 'light client' typical represents.

rvagg:
> the main opinion i'm positing is that the most interesting place to build a light library for deal making & wallet signing is in the browser with JS

:thumbsup: 

I think this then is an evolution of https://github.com/filecoin-shipyard/filecoin.js, which got stalled on #48 from what I understand (was grant funded but this blocker stops it being as powerful as equivalents blockchain universe).

proposals/filecoin-pinning.md
```diff
@@ -0,0 +1,107 @@
...
+_How sure are we that this impact would be realized? Label from [this scale](https://medium.com/@nimay/inside-product-introduction-to-feature-priority-using-ice-impact-confidence-ease-and-gist-5180434e5b15)_.
+
+<!--Explain why this rating-->
+1-3 - we know there should be something. we haven't heard this approach directly asked for.
+
+
+## Project definition
+#### Brief plan of attack
+
+A core JS library that uses the local IPFS library to
```
jnthnvctr, line 57: What would the changes to this flow look like if we were trying to support enabling someone to go from a pinning service to Filecoin? 

A common pattern is developers will pin content in Pinata on behalf of users (common in the NFT space), and so as a user you may not have a local IPFS node running

willscott: I think this depends on the specific flow.
If this is a user with a wallet who is initiating the pin of content already in a pinning service, then a similar design as proposed here could work. Instead of referencing the peer id of the local ipfs node for the miner to pull content from, the content would be available from the pinning service, and the miner could pull it from there, so it essentially would be a couple steps the user's JS library doesn't need to do in that case.

If it is the developer who has previously had the account with the pinning service, and is now wanting to pay for the filecoin storage on behalf of their user as the primary use case we're supporting, we would likely want a different setup. 

Do you have insight into what the more important setup in terms of where the money is coming from we should be thinking about?

proposals/filecoin-pinning.md
```diff
@@ -0,0 +1,107 @@
...
+## Purpose &amp; impact 
+#### Background &amp; intent
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+Currently, to make sure data is persistently stored in a decentralized mechanism as an IPFS user is a complicated process.
+In this position, you need to get access to a filecoin node, either by running a full lotus node, or paying someone else, like
+infura to run one for you. Then, on the lotus command line, you can make a 'deal' with a miner to store data.
+The lotus deal process allows the source of the data to be a local IPFS node, and will then retreave and stream the data
```
vasco-santos, line 17:
```suggestion
The lotus deal process allows the source of the data to be a local IPFS node, and will then retrieve and stream the data
```

proposals/filecoin-pinning.md
jnthnvctr: I think the ideal flow is as you described- so I think the base pattern you're describing makes sense. The main barrier to this being the flow will end up being the cost and whatever other pathways are encouraged in the short term (e.g. batching solutions for small files) -> though I think we should be thinking of that as a separate flow

PR #70, opened 2021-03-05
[WIP] propose: browser nodes can connect to any node out of the box

vasco-santos:
This is still incomplete and in progress

### DIFF
--- a/proposals/browser-nodes-can-connect-to-any-node-out-of-the-box.md
+++ b/proposals/browser-nodes-can-connect-to-any-node-out-of-the-box.md
@@ -1,0 +1,120 @@
+# Browser nodes can connect to any node out of the box
+
+Authors: @vasco-santos
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
+A proposal should contain enough detail for others to understand how this project contributes to our team’s mission of product-market fit
+for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on,
+and any other information relevant for prioritizing this project against others.
+It does not need to describe the work in much detail. Most technical design and planning would take place after a proposal is adopted.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). 
+Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+One of the key environments for the web3 is the browser. Browser nodes have several connectivity challenges that result from their environment, with the most relevant being:
+- unable to listen for incoming dials;
+- no support for TCP;
+- security policies require dials with WSS and DNS addresses.
+
+Circuit relays have been used to work around these limitations. Browser nodes can establish a connection with them and bind to them to receive incoming dials. They can also be used to "translate" a connection from Websockets to TCP and consequently, enable browser nodes to dial GO nodes.
+
+##### Problem Statement
+
+One of the critical pieces in the puzzle is the lack of ability from browser nodes to connect with GO nodes out of the box (majority of the network). By default, GO nodes do not listen on Websockets, nor have DNS + SSL configured. In addition, GO nodes do not bind to relays out of the box. This creates a network fragmentations and results in problems creating pubsub overlay, getting IPNS records, among others.
+
+Having browser nodes to connect with go nodes out of the box essentially means one of:
+- go nodes listen on webrtc by default + we can use limited relays to also negotiate webrtc conns + we deploy nodes that can act as limited relay and have dns + wss addresses (which means not all DHT servers will be capable of doing so out of the box)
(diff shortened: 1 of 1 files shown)


proposals/browser-nodes-can-connect-to-any-node-out-of-the-box.md
```diff
@@ -0,0 +1,120 @@
+# Browser nodes can connect to any node out of the box
+
+Authors: @vasco-santos
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
```
BigLep, line 5:
```suggestion
Initial PR: https://github.com/protocol/web3-dev-team/pull/70
```

proposals/browser-nodes-can-connect-to-any-node-out-of-the-box.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Impact
+_How would this directly contribute to web3 dev stack product-market fit?_
+
+<!--
+Explain how this addresses known challenges or opportunities.
+What awesome potential impact/outcomes/results will we see if we nail this project?
+-->
+
+We had some really good discussions yesterday about the state of Pubsub/IPNS in each runtime environment  + what are the gaps for building dapps on our stack out of the box.
```
BigLep, line 65: Are there notes from the conversation?  I am seeing now that this is WIP.

vasco-santos: Yes, this was a wrong copy paste and not intended for the actual proposal. But here are the notes: https://hackmd.io/@gozala/state-of-ipns

vasco-santos:
Some notes from @lidel that should be important to catalog here: https://github.com/ipfs/in-web-browsers/issues/181

proposals/browser-nodes-can-connect-to-any-node-out-of-the-box.md
```diff
@@ -0,0 +1,120 @@
...
+
+##### Problem Statement
+
+One of the critical pieces in the puzzle is the lack of ability from browser nodes to connect with GO nodes out of the box (majority of the network). By default, GO nodes do not listen on Websockets, nor have DNS + SSL configured. In addition, GO nodes do not bind to relays out of the box. This creates a network fragmentations and results in problems creating pubsub overlay, getting IPNS records, among others.
+
+Having browser nodes to connect with go nodes out of the box essentially means one of:
+- go nodes listen on webrtc by default + we can use limited relays to also negotiate webrtc conns + we deploy nodes that can act as limited relay and have dns + wss addresses (which means not all DHT servers will be capable of doing so out of the box)
+- go nodes listen on websockets address by default + we find a solution for out of the box certificates for announcing dns+wss multiaddr
```
rvagg, line 43: the certificate problem seems like it could be a project in itself, go-ipfs certbot integration or something like that, the whole certificate problem is a diabolical one!

rvagg: Ah, I see https://github.com/ipfs/in-web-browsers/issues/181 touches on some of the issues related to certs, but that doesn't even suggest a clear path forward. That presents us with a scoping challenge here, we have no clear technical solution and this may end up being experimentation to figure out something workable. Ideally we'd have a solid idea to at least try and can scope fail/success with that.

jacobheun: An experimental project here would be to vet the feasibility of a service to issue subdomain certs/DNS mappings to `https://<peerid>.libp2p.io` or similar.

The other option is likely more viable (bind to a limited relay), but, we have to consider that if hole punching fails the nodes won't be able to speak.

Another option here is improving the active relay support story. If browser nodes are able to bind to a local node, or potentially later a relay-as-a-service, then execute dials through the relay as a fallback we can reduce the dependency on the SSL problem. This might look like:
- Improve browser ability to discover local and remote relays to bind to
- Ensure those relays can be/are configured to be active HOP relays (dial on a nodes behalf)
- Improve the prevalence of these relays (more local deploys, ipfs desktop, embedded browser nodes, hosted/incentivized services, electron template app)

rvagg: `https://<peerid>.libp2p.io` sounds like a whole can of worms, similar to the gateway problems (hosting content we don't necessarily want to host) but for more than just content (i.e. someone could abuse the service to host whatever they like, https, on a domain we own. Maybe there's been some thought put into how to deal with these kinds of problems already?

2 other options for managing certificates:

1. Build a certbot plugin to manage certificates for a go-ipfs node, probably with some go-ipfs interaction to setup the challenge process (down side: this is _very_ opt-in, even if we make it super simple for them)
2. go-ipfs already has some user-interactive startup operations that it performs (fs migrations being the main one I think), so another option here is to build in the full letsencrypt certificate management process to node setup - as an optional, but encouraged step. It'd not be a simple project, but there's a full featured go [library](https://github.com/go-acme/lego) for dealing with ACME so we could run the entire process, including renewals, with some initial setup from the user.

The challenge process introduces awkwardness in both options but there could be a simple-path for easy setups (no NAT, no complex firewalls, etc.) and documentation for the more complex-path setups.

vasco-santos: cc @mxinden

mxinden: _Thanks @vasco-santos for the ping._

> Even if we never do the second one, built-in ACME support for setting up /wss on own domain is a good match of real life usage (people self-hosting IPFS infra on own domains), and would make things easier for everyone.

:+1: for option 1 being good enough, especially when "go nodes listen on webrtc by default", as browsers could connect to non-ssl Go nodes via WebRTC with the help of a _smaller_ number of SSL + WSS Go nodes as signaling servers using circuit relay v2.

> but, we have to consider that if hole punching fails the nodes won't be able to speak.

True. Sorry, in case this is a question already discussed: Is the goal of this proposal with `Browser nodes can connect to **any** node out of the box` a hard requirement, or is increasing overall connectivity on a best effort basis good enough?

Asked differently would `Browser nodes can connect to **most** nodes out of the box`, with the remaining nodes being the ones unreachable due to WebRTC NAT hole punching issues, be a goal worth achieving as well? To me 100% connectivity in a peer-to-peer setting seems impossible in the first place, thus I would assume the latter title to be more accurate. What do people think?

PR #71, opened 2021-03-11
propose: js modular crypto

vasco-santos:


proposals/large-crypto-blobs.md
```diff
@@ -0,0 +1,127 @@
...
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives. 
+-->
+
+One of the pain points of development within a browser context is the bundle size of a web app, which easily gets quite big and makes web apps to take longer times loading.
+
+In `js-libp2p` and `js-ipfs`, we have been working towards reducing the bundle size of the library to the minimum possible, so that the users of our stack don't need to ship gigantic bundle size applications just for using parts of our stack. However, we also created technical debt when `libp2p-noise` was added to the stack, as a new crypto library was added.
+
+One of the outstanding issues to have our bundle size in a considerably good shape is the crypto libraries that we need to carry. The JS ecosystem around cryptography is far from perfect, which has made us rely on several different libraries to achieve different cryptographic primitives.
+
+![image info](./js-libp2p-bundle.png)
```
vasco-santos, line 36: image

rvagg:
I think that we can learn from the work we ended up doing in [js-multiformats](https://github.com/multiformats/js-multiformats) which was designed specifically with this problem in mind. It's built for composability and designed to let bundlers do nice things with only including the pieces that you actually use. When consuming multiformats and the newer IPLD stack, you only pull in the narrow stack that you want to use for your application. So if you want dag-cbor + blake2b then you get just the pieces needed for that and strictly nothing else. Which is really nice to use. It's different when bundled in js-ipfs of course but this is all part of the #47 story too.

hugomrdias (approved):
LGTM 
Maybe we can add a note about the big number libs that are duplicated in each of these crypto libs. It's a hard problem to solve but i think we should try to figure it out.

proposals/large-crypto-blobs.md
```diff
@@ -0,0 +1,127 @@
...
+<!--
+Provide success criteria. These might include particular metrics, desired changes in the types of bug reports being filed, desired changes in qualitative user feedback (measured via surveys, etc), etc.
+-->
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
```
Gozala, line 104: - WASM backend was brought up other day, which I think is worth considering in alternatives

Gozala:
One challenge I see which might be worth calling out / articulating in the proposal is how do individual components communicate dependence on availability of certain cryptographic algorithm. Unilke with js-multiformats library it may not be all that obvious what libp2p component may require what.

Gozala:
Are there any ideas how are we going to deal with third party libraries ? My understanding was that quite a few pieces were pulled by dependencies as opposed to libp2p itself.

proposals/large-crypto-blobs.md
```diff
@@ -0,0 +1,127 @@
...
+#### Confidence
+_How sure are we that this impact would be realized? Label from [this scale](https://medium.com/@nimay/inside-product-introduction-to-feature-priority-using-ice-impact-confidence-ease-and-gist-5180434e5b15)_.
+
+<!--Explain why this rating-->
+
+
+## Project definition
+#### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
```
Gozala, line 83: I would suggest  to put crypto interface definition that can be evaluated fairly cheaply as a first step. I think js-multiformats does a good job doing it for [multibase](https://github.com/multiformats/js-multiformats/blob/master/src/bases/interface.ts), [multihash](https://github.com/multiformats/js-multiformats/blob/master/src/hashes/interface.ts), [multicodec](https://github.com/multiformats/js-multiformats/blob/master/src/codecs/interface.ts) and I would love similar thing for crypto.

olizilla:
A tip from @carsonfarmer 
> I'd recommend checking out stablelib for pure js/ts implementations of just about everything crypto: https://github.com/StableLib/stablelib
> – https://github.com/multiformats/js-multiformats/pull/59#issuecomment-769607925

jacobheun:
It looks like leveraging stablelib could help us drastically reduce our dependency footprint. If there are any missing crypto items there that we need for our protocols we could reach out to Dmitry (owner) and fund that work. `libp2p-crypto` could then just become a light, libp2p spec compliant api on top of that.

carsonfarmer:
I actually had started creating a stablelib libp2p-crypto implementation a while back, and should still have the code lying around if anyone wants to take it over. Probably about 50% done? My goal was to simply have a drop-in replacement for libp2p-crypto, which is the biggest culprit for bundle size for me right now. Very much happy to "jam" on any efforts here with the libp2p team, or help contribute if an externally funded project gets going.

I'd be happy to be a point of contact for an external team on the proposal/funded work if that is of use?

hugomrdias:
i found an interesting repo for wasm based hashing https://github.com/Daninet/hash-wasm

proposals/large-crypto-blobs.md
Gozala: - Work with partners (e.g. Igalia) to ship needed crypto primitives natively in browsers.

Gozala:
I have proposed a section to alternatives https://github.com/protocol/web3-dev-team/pull/71#discussion_r599132393. I think it is worth considering what would it take to get necessary crypto APIs in browsers themself as opposed to bundling various user space implementations. It also might be worth looking into crowdfunding options for this as I expect there are plenty of other groups who'd benefit from having necessary crypto APIs in browsers.

mikeal:
There’s a lot of information that has been surfaced since the last edit, I’d like to see it addressed or worked into the proposal before we 👎👍

proposals/large-crypto-blobs.md
rvagg: @Gozala can you give some thoughts here on the timeframe for a usable implementation via this route? This seems like a really good investment to make but we need to be clear about expectations of pay-off time I think.
There's likely cumulative lag involved in the steps that might look something like: (a) identify narrow requirements and negotiate grant, (b) run through spec process to get agreement from browser vendors, (c) implement for (or partner with) browser vendors to get them native .. then (d) wait for releases and critical-mass adoption.

Gozala: > @Gozala can you give some thoughts here on the timeframe for a usable implementation via this route?

I thing it is impossible to make guesses without doing steps below, I'd expect it to be on years scale.

1. Having concrete list of crypto primitives.
1. Having preliminary discussions with browser vendors / standard bodies to asses reception to the idea.
1. Some analyses on the amount of effort it would require across different engines.

Cc @autonome and @lidel who worked with Igalia on protocol registration effort and have some insight on how long it took from idea to having things shipped. 

I also don't think we should not pursue other routes in favor of this, I think it would be wise to do both.

vasco-santos: Added these alternatives. Thanks for raising this @Gozala

vasco-santos:
I added the discussed Alternatives and a brief plan of attack based on the previous discussion of leveraging `stablelib`.

From this and other project discussions, it seems that we should create a project spike on WASM integration to have a better idea of the challenges that we have there, in order to properly compare alternatives

olizilla:
An old map appears! https://github.com/arewedistributedyet/arewedistributedyet/issues/19#issuecomment-384366223 (

2 years ago, the webcrypto guardians were a little more "nope" about streaming APIs but it now sounds like it really needs a champion to guide the work. https://github.com/w3c/webcrypto/issues/73#issuecomment-769602406

wasm crypro libs that do what we need has felt like a useful path to explore for a while. We should support (and sponsor) WebCrypto spec improvements, but to fully unlock that work we'd need to make sure that we could also get browsers to recognise `ipfs://<cid>` as a secure context (@Gozala is that an issue, do you have any insight on that?)

Also an off-topic reminder that https://github.com/arewedistributedyet/arewedistributedyet/issues is a repository of issues that could benefit the dweb beyond IPFS, and if you see anything missing, please add it!

Gozala:
> but to fully unlock that work we'd need to make sure that we could also get browsers to recognise `ipfs://<cid>` as a secure context (@Gozala is that an issue, do you have any insight on that?)

Depends what do you mean exactly. Navigating to ipfs://bafkreic4wacjgff32iqwgip5zolur3dnhiqku2xctmng4vezdqxaxxjkti/ suggests brave does treat it as secure context, while opera on my iphone is not. I'm not sure why Opera arrived to that choice or whether it is an accident, @lidel or @autonome might know more.

As of other browsers today best we can do is [`registerProtocolHandler`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler) which will effectively redirect `ipfs://`. 

Primary challenges when pursuing making ipfs:// a thing at Mozilla had been:

1. The fact that CIDv0 was case sensitive and browsers treat host part of URL as case insensitive. Browsers also don't really do URIs. I think CIDv1 is the most effective way to overcome this challenge.
2.  How does content from ipfs:// interacts with content on https:// and/or http:// is complex and challenging question. I'm unaware anyone tackling this and I think it's going to be a major roadblock in making ipfs:// be more mainstream. If you look at the spec that defines interaction model between https:// and http:// it is massive and really complex, I expect it to be lot more challenging to define interactions with ipfs://. Approach I was going for at Mozilla was not to allow any references across ipfs:// and http(s):// while treating ipfs:// as secure context.

Stebalien (commented):
How much would being able to opt out of some crypto modules actually help? Can we walk through some use-cases, pick the modules we'd be able to drop, then compute the savings?

For example:

* Libp2p probably doesn't need any hash function but blake2 for noise and sha2.
* Libp2p probably does need ed25519, RSA, x509, and sekp (maybe?). That's pretty big... Some applications might not need, e.g., sekp, but they might want to include it anways "just in case".

rvagg:
> Can we walk through some use-cases

I don't quite have the expertise to speak to this, but we should try really hard to extract ourselves from IPFS as it's instantiated today and think about narrow verticals of applications that might consume libp2p (or even just parts of it). What if I want an application that _just_ talks to Filecoin nodes? What about if I have an application that _just_ talks to specific Filecoin nodes (no discovery) purely for the purpose of transferring data for storage or retrieval (pretending graphsync+selectors is cool in JS for a moment)? Or more generically, what if I'm building my own p2p application that _just_ uses a small part of libp2p to throw data back and forth and I have strong opinions on the various protocols that need to align in my stack to make that work.

This proposal should be about unlocking use-cases that currently pull in far too many modules and add overhead and complexity to a build pipeline (including bundling).

jacobheun:

Gozala:
> I have proposed a section to alternatives [#71 (comment)](https://github.com/protocol/web3-dev-team/pull/71#discussion_r599132393). I think it is worth considering what would it take to get necessary crypto APIs in browsers themself as opposed to bundling various user space implementations. It also might be worth looking into crowdfunding options for this as I expect there are plenty of other groups who'd benefit from having necessary crypto APIs in browsers.

@autonome do you think this is something worth having conversation about with igalia ? Seems like something whole web3 space could highly benefit from.

PR #76, opened 2021-03-12
initial proposal of extending gateway to support an external index

willscott:
[Rendered](https://github.com/protocol/web3-dev-team/blob/6cfa3e7834a03446460c2ff75c39b58aa06d564f/proposals/gateway-api.md)

### DIFF
--- a/proposals/gateway-api.md
+++ b/proposals/gateway-api.md
@@ -1,0 +1,94 @@
+# IPFS Gateway can use an external index 
+
+Authors: @willscott
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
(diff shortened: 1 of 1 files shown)


mikeal (approved):

mikeal:
i’d like to see a review approval from someone with more experience working on the gateway, i’m just sure who that would be?

olizilla:
When I worked on IPFS infra last year we tried to manually ensure that all the gateways and cluster nodes were always connected as peers, so they would send out their wantlist to them when a request came in for a CID they didnt already have. The assumption was that a gateway node in another region might have already fetched a popular block, or that we would likely get requests for content that was stored on one of our large storage clusters.

I just found out that go-ipfs config now supports explicit definition of a set of peers https://github.com/ipfs/go-ipfs/blob/master/docs/config.md#peering so that should be much more reliable now.

Could this proposal be achieved by adding these additional indexes as peers?

We could improve all gateways perf by introducing a LRU aware blockstrore, as right now gateways are dumping popular content during GC every few hours. It would benefit desktop and server implementations and content discovery in general if all nodes automatically held on to regularly accessed blocks.

We could radically improve the ipfs.io gateway perf by creating a gloabally shared blockstore on some thing like cloudflare kv.

BigLep:
@mikeal and @olizilla are going to talk verbally about next steps here.

PR #78, opened 2021-03-12
initial proposal for graphsync support in ipfs

willscott:
[Rendered](https://github.com/protocol/web3-dev-team/blob/3b4aa351c14924c7a7423fefd1cb9271851d1a33/proposals/ipfs-graphsync.md)

* [ ] Tagging @hannahhoward to tell me where i'm wrong: in particular if estimation of scope is off, and if i identified reasonable place where this plugs in & challenges.
* [ ] Tagging @aschmahmann  on if there are objections to graphsync being incorporated in this way we should know about

### DIFF
--- a/proposals/ipfs-graphsync.md
+++ b/proposals/ipfs-graphsync.md
@@ -1,0 +1,95 @@
+# IPFS nodes speak Graphsync
+
+Authors: @willscott
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
+A proposal should contain enough detail for others to understand how this project contributes to our team’s mission of product-market fit
+for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on,
+and any other information relevant for prioritizing this project against others.
+It does not need to describe the work in much detail. Most technical design and planning would take place after a proposal is adopted.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). 
+Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+
+Currently, ipfs nodes speak [bitswap](https://docs.ipfs.io/concepts/bitswap/) for exchanging blocks of data. bitswap works at a block layer where you ask for a CID, and get back the block that hashes to that cid.
+
+[graphsync](https://github.com/ipfs/go-graphsync) is a protocol for synchronizing graphs of data across peers. It allows a host to make a single request to a remote peer for all of the results of traversing an IPLD selector on the remote peer's local IPLD graph. It is what filecoin uses.
+
+#### Assumptions &amp; hypotheses
+
+* There are cases where users will want to retrieve a piece of structured data they can efficiently describe with a selector, but which would be expensive to retrieve with bitswap.
+* There are cases where it would be valuable for an IPFS node and a filecoin node to exchange data.
+
+#### User workflow example
+
+* A user would ask for data using a selector in e.g. the [go-fetcher](https://github.com/ipfs/go-fetcher) interface within IPFS (from a plugin)
+* An API is built such that dag exploration of ipld nodes from a client built on IPFS is translated into selectors
+* a client builds a car of data to store, creates a filecoin deal for it, and the miner fetches the data from an IPFS client with graphsync
+
+#### Impact
+
+High. we invision cases where we want IPFS and filecoin nodes to be able to communicate
+
+#### Leverage
+
+High. we have other projects that are enabled by IPFS and filecoin nodes being able to communicate
+
+#### Confidence
+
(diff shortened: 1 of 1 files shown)


proposals/ipfs-graphsync.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### Confidence
+
+Two. We haven't heard direct requests for this flow, but it's our best guess at what's practical.
+
+
+## Project definition
+#### Brief plan of attack
+
+[go-merkledag](https://github.com/ipfs/go-merkledag) currently takes a [blockservice](https://github.com/ipfs/go-merkledag/blob/bf51443272bb98cff071eb44ed9ce6c940e82f1f/merkledag.go#L32).  It should also be able to be created from a graphsync provider, or both. We can initially enable graphsync as an opt-in option, and need to design a transition (e.g. do we have merkledag make requests to both graphsync and to blockservice when present? do we make this code somewhat peer-aware, so we know based on provider which one to use?)
```
aschmahmann, line 57: TLDR: In trying to make GraphSync usable in go-ipfs having an interface that handles both bitswap and graphsync backed DAGServices is the easy part. Content discovery, spreading requests across peers, prioritizing peers + protocols, etc. are the hard parts.

IIUC the main areas GraphSync is lacking in order to be a drop in replacement for Bitswap are:
- We have no code (or AFAIK proposal yet) for distributing data across multiple peers using GraphSync
- Graphsync is not on its own suitable for peer discovery in the way Bitswap is (e.g. it may be less efficient due to being unable to ask for HAVEs)

From what I see this limits our options, and some remaining ones are:
- Make GraphSync able to be a drop-in replacement:
  - Extract sessions code from Bitswap/create a way for sessions to be managed across Bitswap and GraphSync
  - Make sessions smart enough to split queries across peers with GraphSync, or something simpler like trying to use GraphSync only if we fail at using Bitswap for long enough
  -  Use Bitswap (or some other protocol, potentially including GraphSync) for discovery of content from peers we're already connected to
- Have options like `ipfs cat <cid> [--from=<multiaddr>]` that will use GraphSync for direct point-to-point transfers

proposals/ipfs-graphsync.md
```diff
@@ -0,0 +1,95 @@
...
+
+#### User workflow example
+
+* A user would ask for data using a selector in e.g. the [go-fetcher](https://github.com/ipfs/go-fetcher) interface within IPFS (from a plugin)
+* An API is built such that dag exploration of ipld nodes from a client built on IPFS is translated into selectors
+* a client builds a car of data to store, creates a filecoin deal for it, and the miner fetches the data from an IPFS client with graphsync
+
+#### Impact
+
+High. we invision cases where we want IPFS and filecoin nodes to be able to communicate
```
hannahhoward, line 43: If this is the main goal, I'm not sure this proposal is right.

I'm trying to think through this use case: "a client builds a car of data to store, creates a filecoin deal for it, and the miner fetches the data from an IPFS client with graphsync"

Can you explain in more detail what this means? So I have all the data, I make a car file, and then I make a filecoin deal for it -- why don't I just send to the miner with data transfer and graphsync?

What's the use case for the miner fetching with graphsync via an IPFS client? Moreover, why would a miner run a full instance of IPFS when IPFS already supports retrieving via graphsync as an experimental feature -- it seems to me all of the software then is just built into the filecoin mining software-- they don't run a whole IPFS node.

Again, this is a fine project, but as specified the impact is to IPFS data transfer speeds. it really doesn't unlock much on its own for filecoin miners getting data from IPFS.

hannahhoward: also, Filecoin doesn't speak Graphsync: It speaks data transfer on top of graphsync. Just supporting graphsync as a protocol is not enough on its own to unlock Filecoin transfer. IPFS needs to have an integrated data transfer instance to do what I think you want to do here. But again, I am not sure I understand the use case.

hannahhoward (commented):
I'm going to suggest that we have an in depth conversation on goals here, cause I'm not clear what this proposal is intended to unlock, and whether it would in fact unlock the intended capabilities.

As I see it, there are two separate legible projects here:

- IPFS nodes learn to efficiently request data by mixing graphsync and bitswap requests, to improve overall transfer speeds. This is something I have in my head a well defined roadmap and approach for. I've put some components in the comments, but can write it up seperately. 

- Filecoin miners can transfer data for storage deals by requesting it from IPFS. What this requires needs some thought, and is not as clear to me, but I don't think this proposal as described moves us closer to that goal per se. I can see two approaches here:
1. The miner spins up a seperate full IPFS instance to request data. If they do this, there's no need to add graphsync -- just use what IPFS already uses, which is bitswap
2. The miner process itself attempts to talk to IPFS with its existing protocols. If this is the case, graphsync itself is not sufficient -- IPFS would also need to understand data transfer requests.

As I see it, we're mixing a well defined but perhaps lower impact project (IPFS request mixing) with a high priority initiative (integrating IPFS + Filecoin) that it won't actually solve.

BigLep:
Lots of great comments and points raised by @hannahhoward .  When time permits, I assume you'll respond/incorporate @willscott .

willscott:
Yep. This definitely needs a lot more clarity. I think the product asks are still evolving quite a bit as well:
* do we want a go-IPFS node to initiate a connection to a lotus node, or vice versa?
* are we okay speaking bitswap initially?
* what sort of selector support do we need and where does that come into play (e.g. threaded up from an API command being made by a user to the go-ipfs node?)

Decisions on that path (cc @pooja ) will inform the next pass at this project scope.

mikeal:
Can we set this PR to draft until you have time to refine it

willscott:
I don't see a button to do that?

rvagg:
@willscott over just under the reviewer list at the top right ⬈ `Still in progress? Convert to Draft.`

PR #79, opened 2021-03-15
Proposal: TypeScript Definitions for core libraries

rvagg:
This is current work, has been going on for ~1 year now, and is not quite complete. @achingbrain and others have been hard at work trying to wrap it up.

I'm turning it into a project proposal because it has _actual_ high value and high leverage and is a legitimate project that we would very likely prioritise if it wasn't already mostly done. Sudo currently owns this implicitly because it's transitional work, but I'm making it explicit with a project proposal and we're going to properly wrap it up.

@achingbrain, @hugomrdias, @Gozala, @vasco-santos, @jacobheun I'd love some input here to define the right amount of work to complete this in a way that captures as much value as we can and doesn't leave awkward dangling pieces. Are there additional libraries with non-trivial use that are overlooked with the primary focus on js-ipfs dependencies? Is there any additional work needed on tooling (aegir, GitHub Actions, other) that we should have on the list?

Of course I want to keep the scope limited but don't want this to get brushed under the rug because we have real current users who get real value out of this and I think we all expect this work would make it easier to get more users and to increase their utility of our libraries. So let's do it properly, and make sure we can celebrate and get proper acknowledgement for everyone's effort when it's done.

### DIFF
--- a/proposals/79-typescript-definitions.md
+++ b/proposals/79-typescript-definitions.md
@@ -16,7 +16,7 @@ Work on this effort has been largely completed, thanks primarily to @hugomrdias,
 
 Remaining work to integrate the current set of js-ipfs dependencies into js-ipfs and address the typing for code directly in js-ipfs is happening here: https://github.com/ipfs/js-ipfs/pull/3550
 
-Aside from completing the remaining js-ipfs itegration work, the scope of this project includes some additional libraries that re not currently part of the js-ipfs dependency tree, including:
+Aside from completing the remaining js-ipfs integration work, the scope of this project includes some additional libraries that are not currently part of the js-ipfs dependency tree, including:
 
  * Next-generation IPLD codec libraries (using the js-multiformats pattern)
  * _TODO: what other non-archived, non-dormant project should we include here to achieve the above value & impact ideals?_
--- a/proposals/79-typescript-definitions.md
+++ b/proposals/79-typescript-definitions.md
@@ -1,0 +1,54 @@
+# TypeScript Definitions for core libraries
+
+Authors: @rvagg
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/79
+
+## Purpose &amp; impact 
+
+Consumers of our JavaScript libraries & components should have sufficient TypeScript definitions, available through standard means, to write fully typed TypeScript code, or use that code to drive tooling that consumes these definitions - such as VS Code editing enhancements, documentation production pipelines, type checking tooling for test & CI pipelines.
+
+This work is high-value and high-impact for the JavaScript ecosystem and those of us working on open source JavaScript libraries can all provide anecdotal evidence for the frequency with which developers request better TypeScript annotations for our libraries. The rate of TypeScript adoption, particularly within larger-scale projects, is increasing, but TypeScript definitions are increasingly useful as they are included in general JavaScript linting, checking and documentation tooling.
+
+This work provides high-leverage within our suite of JavaScript tools as we mature, refactor, modularize and create. We are already establishing a suite of practices and tooling that are used in varying ways across PL JavaScript projects that use definitions, even though we have very few TypeScript projects throughout our GitHub orgs (<https://github.com/ipfs/js-dag-service> being a rare example, which was initially contributed by Textile). It is reasonable to expect that the majority of new JavaScript code produced by Protocol Labs into the future will make use of TypeScript annotations in some way.
+
+Work on this effort has been largely completed, thanks primarily to @hugomrdias, @Gozala, and @achingbrain with assistance from the entire JavaScript team in various ways. The bulk of the activity has been tracked as a roll-up in to js-ipfs @ <https://github.com/ipfs/js-ipfs/issues/2945>.
+
+Remaining work to integrate the current set of js-ipfs dependencies into js-ipfs and address the typing for code directly in js-ipfs is happening here: https://github.com/ipfs/js-ipfs/pull/3550
+
+Aside from completing the remaining js-ipfs itegration work, the scope of this project includes some additional libraries that re not currently part of the js-ipfs dependency tree, including:
+
+ * Next-generation IPLD codec libraries (using the js-multiformats pattern)
+ * _TODO: what other non-archived, non-dormant project should we include here to achieve the above value & impact ideals?_
+
+Project-specific decisions will be made regarding the depth of TypeScript definitional work. Projects with greater expected future usage should include full type checking in CI and will therefore require basic inline TypeScript annotations. Projects that are dependencies but are not expected to be actively maintained or developed further into the future may just include basic API type definitions so that dependents can make use of those.
+
+<!--
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+#### User workflow example
+_How would a developer or user use this new capability?_
+#### Impact
+_How would this directly contribute to web3 dev stack product-market fit?_
+#### Leverage
+_How much would nailing this project improve our knowledge and ability to execute future projects?_
+#### Confidence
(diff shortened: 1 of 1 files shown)


proposals/79-typescript-definitions.md
```diff
@@ -0,0 +1,54 @@
...
+
+Work on this effort has been largely completed, thanks primarily to @hugomrdias, @Gozala, and @achingbrain with assistance from the entire JavaScript team in various ways. The bulk of the activity has been tracked as a roll-up in to js-ipfs @ <https://github.com/ipfs/js-ipfs/issues/2945>.
+
+Remaining work to integrate the current set of js-ipfs dependencies into js-ipfs and address the typing for code directly in js-ipfs is happening here: https://github.com/ipfs/js-ipfs/pull/3550
+
+Aside from completing the remaining js-ipfs itegration work, the scope of this project includes some additional libraries that re not currently part of the js-ipfs dependency tree, including:
+
+ * Next-generation IPLD codec libraries (using the js-multiformats pattern)
+ * _TODO: what other non-archived, non-dormant project should we include here to achieve the above value & impact ideals?_
```
jacobheun, line 22: Tooling is mentioned above, but we should clarify what we aim to land as part of this. Quite a bit of work has gone into Aegir for type generation, what's the short term milestone/feature set we're targeting for that? Is it already done?

proposals/79-typescript-definitions.md
```diff
@@ -0,0 +1,54 @@
...
+
+Work on this effort has been largely completed, thanks primarily to @hugomrdias, @Gozala, and @achingbrain with assistance from the entire JavaScript team in various ways. The bulk of the activity has been tracked as a roll-up in to js-ipfs @ <https://github.com/ipfs/js-ipfs/issues/2945>.
+
+Remaining work to integrate the current set of js-ipfs dependencies into js-ipfs and address the typing for code directly in js-ipfs is happening here: https://github.com/ipfs/js-ipfs/pull/3550
+
+Aside from completing the remaining js-ipfs itegration work, the scope of this project includes some additional libraries that re not currently part of the js-ipfs dependency tree, including:
```
jacobheun, line 19:
```suggestion
Aside from completing the remaining js-ipfs integration work, the scope of this project includes some additional libraries that are not currently part of the js-ipfs dependency tree, including:
```

proposals/79-typescript-definitions.md
```diff
@@ -0,0 +1,54 @@
...
+
+ * Next-generation IPLD codec libraries (using the js-multiformats pattern)
```
vasco-santos, line 21: Does this include multiaddr update? https://github.com/multiformats/js-multiaddr/pull/159

proposals/79-typescript-definitions.md
rvagg: I'm interested in input on this, maybe from @hugomrdias or @achingbrain? The aim is something like "get it done properly" but also "not too large in scope that it goes beyond a S/M project". So I don't want to blow this out with tooling investment.

proposals/79-typescript-definitions.md
rvagg: I think it probably should be, but I don't have a feel for how much work it is to get that over the line. Do you?

vasco-santos: maybe @hugomrdias have an idea on the current state. More than having that merged, this will require changes across the stack as this will change how multiaddrs are created

momack2:
https://github.com/libp2p/js-libp2p/issues/830

proposals/79-typescript-definitions.md
hugomrdias: That PR should be made non-breaking just to get it finished, i tried a week ago or something but its troublesome. Need to look at it again.

proposals/79-typescript-definitions.md
hugomrdias: Regarding types aegir is pretty much done.

hugomrdias: Theirs another angle regarding types, which is documentation. Aegir uses typedoc and it does a pretty good job at taking our types and generating documentation.
But because our code base is still commonjs there is still some minor issues. To overcome these issues i normally need to either change the current commonjs exports to something more friendly or file issues in Typedoc to support all the quirks about commonjs exports :)
This may be out of scope but still, up to date documentation is important!

rvagg:
In f1e6f2eI've tried to tighten up the scope of this project. It's important that this not become a catch-all for anything TS related and we actually have an end-point. I've also tried to fill out a lot of the missing parts of the project.

@anorth @jacobheun I'd really appreciate some reviews and feedback regarding how I've tried to define scope and some of the additional pieces. As noted there is risk of scope creep in this one and I'm keen to extract maximal value for this without it growing beyond the very rough estimates available at the moment.

proposals/79-typescript-definitions.md
rvagg: @hugomrdias :thumbsup: my take is that the tooling is not mature enough for the annotated-JS case to do a consistently good job of it--it doesn't seem to be a priority on the typedoc end and doing it from the jsdoc end (as I've been trying to do) just doesn't work the more you lean in to TS (nobody seems to be interested in unifying the two). But this is certainly something we should take on in a near-future project, to build better and more consistent docs across our stack with tight integration with the TS annotations.

rvagg:
latest (force pushed) commit is f0bc4c5 that has all the changes

rvagg:
urgh, f1e6f2e, sorry, mutable content addressing is hard, who knew?

jacobheun:
For scope, do we have a clear target that's driving the extra typing outside of completing the existing js-ipfs work? Right now, it's not clear to me what's driving the ipld-codec or multiformats work, although I have assumptions about why. Perhaps it would be useful to identify gaps in downstream projects like https://github.com/ceramicnetwork/js-ceramic? I briefly looked through the code and noticed there aren't a ton of custom types in the repo, but there are several places where ipfs calls are `@ts-ignored`'d. As one of the primary goals of this is for downstream type leverage, would selecting a couple of these larger projects help to validate/narrow our scope?

proposals/79-typescript-definitions.md
```diff
@@ -0,0 +1,93 @@
...
+
+Anecdotally, developers find type annotations useful in the development process even if they don't use TypeScript. It also allows for additional checking in the test/CI process for dependents of our libraries.
+
+#### Leverage
+_How much would nailing this project improve our knowledge and ability to execute future projects?_
+
+We have some existing TypeScript in the PL suite of JavaScript. This may expand as we have some developers who have a preference for working with strongly typed, or minimally type checked code. Annotations on our core libraries allows us to add tighter type checking to our CI process for any existing and new libraries that consume that code.
+
+We already have [one example](https://github.com/ipfs/js-dag-service) of code being contributed to the PL stack from a third-party that uses TypeScript and could be improved by typing in the rest of our stack.
+
```
Gozala, line 52: I think ease of onboarding new contributors is worth calling out specifically. When I joined, I found amount of code so overwhelming that I end up writing some type annotations so I could leverage tools to see what was passed around as opposed to trying to keep it all in my head.

Gozala: Also worth calling out that in the process of entyping js-ipfs and it's dependencies we have identified and fixed fare amount of otherwise unknown bugs in the code base.

proposals/79-typescript-definitions.md
Gozala: I'm not sure where exactly to fit this and I believe it is out of scope for this specific proposal, but thought I'd still call this out here. We leverage TS a bit across message-port-client / server pieces to ensure that both ends comply with a wire protocol, I think it could be generalized a bit more to provide same value across more places where we hav client / server or other RPC model.

rvagg:
Address comments in latest commit:

* why the IPLD libraries are included (although I consider them less critical path here and they could be scrapped if we can't solve some of the issues as easily as I hope)
* added some of @Gozala's comments re leverage/impact of this work
* attempted to add success metrics by looking at downstream users, thanks to @jacobheun's suggestion

jacobheun (approved):

hugomrdias (approved):

PR #82, opened 2021-03-17
propose: js projects configuration

vasco-santos:
This project aims at improving the JS projects configuration, which has been one of the biggest barriers for users to get up to speed in our stack. The challenge here is defining the scope or decoupling this project into multiple sub-projects. I have a big context on the libp2p side of things, but would like to get input from other folks to include the remaining projects.

### DIFF
--- a/proposals/82-js-configuration.md
+++ b/proposals/82-js-configuration.md
@@ -1,0 +1,120 @@
+# [outcome or objective here] 
+
+Authors: @vasco-santos <!-- List authors' GitHub or other handles -->
+
+Initial PR: https://github.com/protocol/web3-dev-team/pull/82 <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
+This template is for a proposal/brief/pitch for a significant project to be undertaken by a Web3 Dev project team.
+The goal of project proposals is to help us decide which work to take on, which things are more valuable than other things.
+-->
+<!--
+A proposal should contain enough detail for others to understand how this project contributes to our team’s mission of product-market fit
+for our unified stack of protocols, what is included in scope of the project, where to get started if a project team were to take this on,
+and any other information relevant for prioritizing this project against others.
+It does not need to describe the work in much detail. Most technical design and planning would take place after a proposal is adopted.
+Good project scope aims for ~3-5 engineers for 1-3 months (though feel free to suggest larger-scoped projects anyway). 
+Projects do not include regular day-to-day maintenance and improvement work, e.g. on testing, tooling, validation, code clarity, refactors for future capability, etc.
+-->
+<!--
+For ease of discussion in PRs, consider breaking lines after every sentence or long phrase.
+-->
+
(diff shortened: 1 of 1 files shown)


proposals/82-js-configuration.md
```diff
@@ -0,0 +1,120 @@
...
+#### Brief plan of attack
+
+<!--Briefly describe the milestones/steps/work needed for this project-->
+
+- We have been cataloging in libp2p all the problems of configuration we see from issues and have some discussions about potential solutions in https://github.com/libp2p/js-libp2p/issues/576. But the scope is much broader than libp2p
+
+#### What does done look like?
+_What specific deliverables should completed to consider this project done?_
+
+TODO, we need to define the scope
```
rvagg, line 84: Perhaps we can deal with the scoping challenge by limiting it to the libp2p work as round-1 to form stronger opinions about how to do this well and then do subsequent iterations/projects doing that elsewhere. The proposal in https://github.com/libp2p/js-libp2p/issues/576#issuecomment-594590865 looks like a pretty solid suggestion for how to move forward and we could just adopt that as the basis of a plan (with updates to account for newer perspectives).
There's probably some overlap with #71 and some of the work we did in multiformats/ipld (e.g. see https://github.com/multiformats/js-multiformats/blob/master/src/basics.js for a "give me the common basics" version vs the "I'll build my own, thanks"). I imagine there's opportunities here to decouple code paths so that, in configuring your libp2p you're instantiating a set of dependent modules (& packages) that can be easily tree-shaken by a bundler.

jacobheun: I think limiting to libp2p initially would be beneficial, it's the most challenging to configure at this point. A possible amendment, although it could be done as a followup, would be to setup deployment to automatically release the various configurations as ready to go libp2p modules, similar to https://github.com/carsonfarmer/libp2p-bundle, which would reduce the need to install any other modules to get up and running. The multiformats/ipld work is a great reference for potential structure here. If we built this out, we could later go through crypto and shake out the needed components for each of these bundles, namely browsers, to reduce size for those configs.

vasco-santos: I agree with decoupling this into libp2p first as I think that is where most of the pain points are.

However, there are some issues on `ipfs` configuration "conflicts" with libp2p per https://github.com/ipfs/js-ipfs/issues/3590#issuecomment-803509950, where we have multiple ways of configure things and then surprising end results might appear.  This can probably be a maintenance task in ipfs or something that the onboarding team can have a look. What do you think?

jacobheun: > This can probably be a maintenance task in ipfs

That or it could eventually become a project. To keep the scope small here, starting with making libp2p easy to construct and then have a default build for Node.js and the browser that can be used on the public network out of the box would be a good starting point. This gets immediate value to new users so they can avoid configuration, and makes it easier for us to roll these changes into js-ipfs and improve its configuration later.

jacobheun:

PR #86, opened 2021-03-26
Download migrations over IPFS

Stebalien:


proposals/migrate-over-ipfs.md
```diff
@@ -0,0 +1,152 @@
...
+
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+Users can start a new version of go-ipfs that switches to a new repo version without downloading anything from a centralized (easy to block) server.
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+1. Downloading migrations over IPFS could be too slow/unreliable. Falling back on a gateway should mitigate.
```
lidel, line 105: To make dogfooding safer, we could initially set implicit default to keep HTTPS attempt as the first one and use IPFS as a fallback (for China etc):

```
["dweb.link", "ipfs"]
```
This way the default behavior does not change for most users (migration does not take long time), and we fix edge cases safely.

proposals/migrate-over-ipfs.md
```diff
@@ -0,0 +1,152 @@
...
+
+```js
+{
+    "Migration": {
+        // When true, always use IPFS. When false, never use IPFS. When unset, pick the default
+        // behavior.
+        "UseIPFS": true|false,
+        // When true or unset, use the default gateway. When false, don't use a gateway. When a
+        // string, use the specified gateway.
+        "UseGateway": true|false|"my-gateway",
```
lidel, line 78: I feel this could be simpler and more flexible, how about something like:

```suggestion
        // Control how migration should be downloaded.
        // "ipfs" means fetching over IPFS, everything else indicates the name of HTTP gateway.
        // When left empty, an implicit default of ["ipfs", "dweb.link", "ipfs.io"] is used.
        "DownloadPolicy": []
```

lidel (approved):
We _really_ need this.  The current state is too brittle for use in Desktop / Brave.

Breaking updates are [long running frustratiin to our Desktop  users](https://github.com/ipfs/ipfs-desktop/issues?q=is%3Aissue+fs-repo-migrations), and we should avoid exposing Brave users to the same problem.

proposals/migrate-over-ipfs.md
Stebalien: Fair enough.

proposals/migrate-over-ipfs.md
Stebalien: Hm. Yeah, that works. I'd still like a "default-gateway" option but this is nice and flexible (and allows the user to specify the order).

proposals/migrate-over-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+    2. If `Keep` is "pin", pin the migration binary.
+7. If running, shutdown any temporary nodes and delete the temporary repo.
+
+**Implementation steps:**
+
+1. Implement everything but step 6.
+2. Implement step 6. Unless nodes keep around a copy of the migration, this feature isn't going to be all that useful.
+3. Eventually, switch over to using IPFS by default instead of the gateway as the first option.
+
+NOTE: the only reason we're proposing using the gateway first is that the migration blocks startup so it needs to be fast and reliable.
```
aschmahmann, line 100: What if we set up peering to the relevant gateway or cluster nodes so we defaulted to IPFS and dogfood Bitswap? On the surface this seems easy, is it worthwhile?

proposals/migrate-over-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+####  What does success look like?
+_Success means impact. How will we know we did the right thing?_
+
+Users can start a new version of go-ipfs that switches to a new repo version without downloading anything from a centralized (easy to block) server.
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+1. We may not be able to download/find the migration on IPFS. Trying the gateway first (or racing) should help mitigate this.
+2. The "temporary" node may be missing important parts of the configuration. For example, it may need alternative bootstrap nodes as our nodes may not be reachable. We'll need to think about this carefully and may need to "reach" into the old repo's config file a bit.
```
aschmahmann, line 117: Could totally be a problem with a few possible ways to fix it, it might not be so bad to wait for the user reports to come in before deciding how to resolve this.

One perhaps simple way to deal with this would be allowing users to pass a flag `ipfs daemon --migration-config=/path/to/config` and then load that config file into the temporary node. It's a bit of a hassle for the user, but would certainly handle the job.

proposals/migrate-over-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+##### Domain Fronting
+
+A very simple alternative is to simply use domain fronting. If we registered, e.g., ipfs-dist.com and made it an alias for dist.ipfs.io, it _might_ work in China.
+
+However, we'd still be using HTTPS instead of IPFS to download migrations, which seems kind of silly.
+
+##### Bundle the migrations
+
+We could bundle all migrations with go-ipfs. But that would add 100s of megabytes to the go-ipfs distribution so we really don't want to do this.
+
```
aschmahmann, line 131: Another alternative, which we'd like to do but is probably more work, is to have a system we trust enough to distribute binaries that are auto-executed by the user.

This would enable us to do auto-updating of go-ipfs and cover migrations at the same time. The downside here is that it's a) more work b) requires a bit (although not necessarily a ton) of design on how we to do this securely.

aschmahmann (approved):
Left a few comments, but overall 💯

proposals/migrate-over-ipfs.md
Stebalien: I'd like to _start_ with HTTPs by default so we can get this in ASAP. Then, maybe in the next release, switch to IPFS by default so we have some time to test.

But yeah, peering by default sounds like a great idea.

Stebalien: @aschmahmann could you add that to the proposal?

BigLep:
I'm not sure the right assignee for this since we don't have our project teams modeled in the Protocol organization, but this is getting owned by the "Data Systems" team.

proposals/migrate-over-ipfs.md
```diff
@@ -0,0 +1,163 @@
...
+```js
+{
+    "Migration": {
+        // Sources in order of preference where "HTTPS" means our gateways and "IPFS" means over IPFS.
+        // (an empty list means "do the default thing")
+        "DownloadPolicy": ["HTTPS", "custom-gateway.io", "IPFS"],
+        // Whether or not to keep the migration after downloading it.
+        "Keep": "pin"|"cache"|"discard"
+    }
+}
```
gammazero, line 70: Should this be read from the existing config file, or something that a user puts in a separate migration config file?  The existing config file may not be readable until after migration is performed.

What about having this specified by an optional CLI flag?
```
--dlpolicy="https,ipfs,custom-gateway.io,..." 
```

Stebalien: GAHHHHHHH. This is so, very, annoying.

What if we _just_ decoded this part of the config? As long as this part of the config remains somewhat consistent across upgrades (and we correctly handle missing fields), we should be fine.

Basically:

1. Get the path to the repo.
2. Read the config file into a `struct { Migrations config.MigrationConfig }` type.
3. Move on.

This is significantly safer than, say, reading the swarm config, because it _only_ deals with migrations.

Thoughts?

gammazero: That is fine. Will go with config file.

gammazero: I think it should also include:
```go
    // Peers lists the nodes to attempt to connect with when downloading migrations.                                                                     
    Peers []peer.AddrInfo                                                                                                           
```

PR #88, opened 2021-03-26
template: improve the confidence section

Stebalien:
Currently, it asks:

> How sure are we that this impact will be realized?

Where "impact" is:

> How would this directly contribute to web3 dev stack product-market fit?

I read this as:

> How sure are we that this project will successfully achieve the
"impact" specified in the impact section?

However, I don't believe this was the intended question. The intention was to determine how confident we are that we're solving the _right_ problem, not how confident we are that the proposal will solve the problem.

I've changed the template to ask:

> How sure are we that this is the right problem to work on?

### DIFF
--- a/proposals/TEMPLATE.md
+++ b/proposals/TEMPLATE.md
@@ -51,7 +51,7 @@ Explain the opportunity or leverage point for our subsequent velocity/impact (e.
 -->
 
 #### Confidence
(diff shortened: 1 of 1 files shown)


BigLep:
Your updated question makes sense.  

Is this a case where we should be asking both questions: 
1. How sure are we that this project will successfully achieve the "impact" specified in the impact section?
2. How sure are we that this is the right problem to work on?

rvagg:
> Is this a case where we should be asking both questions:

It'd be nice if we could resist adding more questions. There's already too many questions with no clear pipeline from the answers to their actual use. If you want to add more questions, can you be clear about how their answers will contribute to project selection?

PR #90, opened 2021-03-27
Add Issue Labels

Stebalien:
These issue labels were imported from https://github.com/ipfs/community/raw/master/ISSUE_LABELS.md (with slight modifications) which was originally adapted from https://github.com/libp2p/libp2p/blob/master/ISSUE-LABELLING.md.

- The `status/duplicate` label was dropped. This is really a resolution, not a status. Duplicate issues are closed with a comment anyways so this label doesn't help much.
- The `status/deferred` label was dropped. Priorities should be communicated through the `P*` labels. Any `P2` & `P3` issues are effectively "deferred".
- The "difficulty" labels were renamed to "exp" for expertise to match how we use them. These labels now communicate the amount of experience with a system/repo required to tackle the issue.
- I've added a few more topic labels, but most of these will be defined on a per-repo basis.

This patch _also_ contains a GitHub Action to verify that all issue labels have been updated upstream in the [protocol/.github](https://github.com/protocol/.github) repo. In the future, I'd like to partially automate this update process. Unfortunately, handling renames is a bit tricky so we're going to have to keep these two lists in sync for now.


Stebalien:
I guess the obvious question is: is this the _right_ repo for this label taxonomy? We need _a_ consistent place for it given that we're now a single team working across 5 different orgs.

proposals/dev-network-testsuite.md
```diff
@@ -0,0 +1,104 @@
+# Automated integration testing & test vector extraction
```
BigLep, line 1: I assume this shouldn't be in this PR.

ISSUE_LABELS.md
```diff
@@ -0,0 +1,120 @@
...
+### Kind
+Overarching type of issue or PR. For an additional layer of specificity, use the `area` label.
+
+| Label               | Description                                                             | Color     |
+| -----               | -----------                                                             | -----     |
+| `kind/architecture` | Core architecture of project                                            | `#c7def8` |
+| `kind/bug`          | A bug in existing code (including security flaws)                       | `#fc2929` |
+| `kind/discussion`   | Topical discussion; usually not changes to codebase                     | `#c7def8` |
+| `kind/enhancement`  | A net-new feature or improvement to an existing feature                 | `#c7def8` |
+| `kind/maintenance`  | Work required to avoid breaking changes or harm to project's status quo | `#c7def8` |
```
BigLep, line 37: Were we going to add a note to be more specific here?  Maybe something about "Build systems, dependency upgrades, CI."

BigLep (approved):
Awesome - great work!

proposals/dev-network-testsuite.md
Stebalien: lol. Thanks for catching.

ISSUE_LABELS.md
Stebalien: Good point. The current description better describes "tech debt" which probably deserves a different label.

Stebalien: It looks like we need to split this into:

- `kind/maintenance`: CI, labeling, permissions.
- `kind/cleanup` (or `kind/debt`): Cleanup, refactors, etc.

mikeal:
Most of these labels don’t apply well to a repo that isn’t code and having a long list of unused labels will make labelling more difficult.

Stebalien:
> Most of these labels don’t apply well to a repo that isn’t code and having a long list of unused labels will make labelling more difficult.

For now, I'm happy to apply these labels to code repos only (by default).

ISSUE_LABELS.md
BigLep: Sounds good to me!

BigLep (approved):

PR #92, opened 2021-03-29
proposal: Minty in the browser as a decentralized web app

yusefnapora:
This adds a proposal to port Minty from a node.js command line app to something that can run completely in the browser, using js-ipfs and an ethereum wallet extension like MetaMask.

I'm not 100% convinced we should start working on this yet, but I wanted to write up the path while Minty is fresh in my head.

### DIFF
--- a/proposals/minty-in-the-browser.md
+++ b/proposals/minty-in-the-browser.md
@@ -1,0 +1,159 @@
+# Minty in the browser - Mint NFTs end to end inside a decentralized web app using js-ipfs
+
+Authors: @yusefnapora
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+
+## Purpose &amp; impact 
+#### Background &amp; intent
+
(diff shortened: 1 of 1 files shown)


proposals/minty-in-the-browser.md
```diff
@@ -0,0 +1,159 @@
...
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+
+- D-app developers want to mint NFTs using IPFS without relying on a centralized backend server
+  - This may not actually be the case. I'm not aware of any NFT platform today that's "fully decentralized" in the sense of doing everything on chain without a backend API.
+
+- We want to encourage d-app devs to use js-ipfs directly in their d-apps, as opposed to using an HTTP api from Pinata, nft.storage, etc
+  - This seems like a good thing to encourage, but it also raises the bar for adoption a bit, since js-ipfs is a heavier dependency than an HTTP client and may require learning new concepts, etc.
+  
+- D-app devs are willing to use a remote pinning service for persistence
```
momack2, line 24: how different is the experience for a dapp creator if they use a remote backend server (a la infura) or a remote pinning service (a la pinata)? I guess the difference is using an HTTP API vs using IPFS directly to parse/upload the file client side?

yusefnapora: I don't think it's a huge difference, since you'll probably want to use a pinning service either way.

If you were running a backend server, you could potentially pin your data to your own IPFS nodes instead of using a pinning service, as long as those nodes are well-connected and up all the time. But since browser nodes can't really be up all the time, it seems like the pinning service is closer to "required" than "recommended".

There's probably a way to do it without a pinning service that I'm not thinking of though... hopefully someone will think of it :)

PR #93, opened 2021-03-29
propose normalizing addressing

Gozala:


proposals/normalize-addressing.md
```diff
@@ -0,0 +1,175 @@
...
+This proposal suggests to address those pain points through following:
+
+1. Making `ipfs://${CIDv1}` canonical representation.
+   > 
+   > We can drive adoption by adopting these representation across our libraries [like propsed here](https://github.com/multiformats/js-multiformats/issues/71)
+   > 
+
+2. Ensure that users see `ipfs://${CIDv1}` regardless of gateway they're on.
+   > Above examples from 2nd to 5th could redirect to:
+   > `ipfs://${originHash(CID)}.dweb.link/ipfs://${CIDv1}/`
```
lidel, line 62: No cigar. Subdomains are how we want people to load websites because content root is aligned with URL root.
Adding this artificial path prefix breaks that alignment and breaks relative paths on every website :(

proposals/normalize-addressing.md
```diff
@@ -0,0 +1,175 @@
...
+- Increased understanding that `gateway` in `http://${gateway}/ipfs://${CIDv1}` is a user choice.
+
+
+
+#### Counterpoints &amp; pre-mortem
+_Why might this project be lower impact than expected? How could this project fail to complete, or fail to be successful?_
+
+- IPFS aspired to replace URL addressing with path addressing which this proposal is at odds with.
+- Despite our efforts for avarage web user `ipfs://` style links are not clickable so it does not pick up.
+
```
lidel, line 144:
```suggestion
- Adding yet another notation does not solve the problem, but magnifies it.
- Changing the default paths complicate setup on reverse proxies.
- Having `://` twice in URL will make IPFS links look even more suspicious (less technical users) or simply wacky (technical) hurting us more than helping.

```

lidel (changes requested):
I am ok with adding support for `https://${gateway}/ipfs://${CIDv1}` notation on gateways as a cheap way of improving UX, but it should be a convenience redirect to a URL following an existing path or subdomain convention:

```
https://${gateway}/ipfs://${CIDv1}
  → https://${gateway}/ipfs/${CIDv1}
    → https://${CIDv1}.ipfs.${gateway}
```

We can't make it a canonical notation because its not compatible with  subdomain root requirement (details below)

Gozala:
> I am ok with adding support for `https://${gateway}/ipfs://${CIDv1}` notation on gateways as a cheap way of improving UX, but it should be a convenience redirect to a URL following an existing path or subdomain convention:
> 
> ```
> https://${gateway}/ipfs://${CIDv1}
>   → https://${gateway}/ipfs/${CIDv1}
>     → https://${CIDv1}.ipfs.${gateway}
> ```
> 
> We can't make it a canonical notation because its not compatible with subdomain root requirement (details below)

What I was suggesting here it should do following instead

```
https://${gateway}/ipfs://${CIDv1}
 →https://${CIDv1}.ipfs.${gateway}/ipfs://${CIDv1}
```

The whole point is that people can ignore the domain and will always see `ipfs://${CIDv1}` making whole pattern recognition work.

proposals/normalize-addressing.md
Gozala: > No cigar. Subdomains are how we want people to load websites because content root is aligned with URL root.

My understanding is we want to make sure that content is origin isolated based on CID. We can accomplish that without teaching people how to translate CIDs into origins etc... It is a lot easier to teach programs like gateways to do that, for humans we could just keep `ipfs://${CIDv1}/` regardless of (sub)domain. That gives us:

1. Origin base content separation we seek.
2. We retain same path in form of `ipfs://${CIDv1}` across all gateways. Reinforcing  that `${bridge}` in `${bridge}ipfs://${CIDv1}` is just a bridge to `ipfs://${CIDv1}`.

> Adding this artificial path prefix breaks that alignment and breaks relative paths on every website :(

You mean sites that use `/images/cat.png` would be broken ? That is a fare point, although I think something that we can overcome by redirecting to `/ipfs://{CIDv1}/images/cat.png` based on referrer.

Also would not those sites be as broken when loaded from `https://ipfs.io/ipfs/${CIDv1}` now ?

lidel: Correct. We spent years fixing bugs in various static website generators to support truly relative paths, and it remained a big adoption blocker because there was always a new bug for something new filled.

Only when subdomain gateways got introduced, and the CID root became the root of URL path, all websites and tools started working out-of-the-box with IPFS. 

:point_right:  Redirects are transparent in browser context, so your fix will work for a subset of problems like loading images or other assets, but it won't fix websites that execute JS code based on  `window.location.pathname`. 

Due to this I don't believe this is feasible.

proposals/normalize-addressing.md
```diff
@@ -0,0 +1,178 @@
+# Normalize ipfs://${CIDv1}
```
Gozala, line 1:
```suggestion
# ipfs:// everywhere
```

proposals/normalize-addressing.md
Gozala: > 👉 Redirects are transparent in browser context, so your fix will work for a subset of problems like loading images or other assets, but it won't fix websites that execute JS code based on `window.location.pathname` 💔

That historical context is really helpful! Do you by a chance have a pointers to above JS problems, so it could be quantified and pros and cons could be evaluated.

I realize this is an unfortunate tradeoff. Then I also think that turning `ipfs://{cid}/path/file` into `https:///${cid}.gateway.tld/path/file` is a cause of why later form is more prominent. I also believe that later form makes it really difficult for an average person to realize that gateway could be swapped.

Maybe there is a compromise, that make better tradeoffs ?

PR #97, opened 2021-04-12
proposal: Lotus dynamic retrieval pricing.

raulk:


proposals/lotus-dynamic-retrieval-pricing.md
```diff
@@ -0,0 +1,142 @@
...
+#### Background &amp; intent
+_Describe the desired state of the world after this project? Why does that matter?_
+
+A retrieval client asks the miner for a quote before proceeding to retrieve a
+piece or a part therein. The miner quotes:
+
+- minimum price per byte
+- maximum payment interval
+- maximum payment interval increase
+- unseal price
```
dirkmc, line 17:
```suggestion
- unseal price
  The price the client must pay if the miner needs to unseal data
- minimum price per byte
  The price the client must pay for each byte of data
- maximum payment interval
  The number of bytes that the provider will send before pausing the transfer to wait for payment
- maximum payment interval increase
  The increase in the payment interval each time payment is received
```

proposals/lotus-dynamic-retrieval-pricing.md
```diff
@@ -0,0 +1,142 @@
...
+
+The quoted values become part of the deal proposal. Unlike with storage deals,
+retrieval deal proposals don’t go on chain.	
+
+All these values are filled from the “stored ask”. The “stored ask” can be
+changed through JSON-RPC, but it’s statically set. Lotus doesn’t have the
+ability to change the quote based on the piece that’s being requested.
```
dirkmc, line 24:
```suggestion
On the provider side, these values are filled from the “stored ask”. The “stored ask” can be
changed through JSON-RPC, but it’s statically set. Lotus doesn’t have the
ability to change the quote based on the piece that’s being requested.
```

proposals/lotus-dynamic-retrieval-pricing.md
```diff
@@ -0,0 +1,142 @@
...
+"stored ask" will be a blanket policy for all retrievals, fast or ordinary,
+verified or unverified.
+
+Impact of the blanket policy to unsealing price:
+- If the unsealing price is set to zero, the miner commits to serving all deals
+  at zero cost, even those that require unsealing. Unsealing is a heavy and
+  expensive operations, one that's irrational for miners to offer for free.
+- If the unsealing price is set to non-zero, fast retrievals will require the
+  creation of a payment channel, which introduces significant latency and
+  requires chain interaction.
```
dirkmc, line 37:
```suggestion
- If the unsealing price is set to zero, the miner commits to serving all deals
  without requiring an unseal payment from the client, even those that require unsealing.
  Unsealing is a heavy and expensive operation, one that's irrational for miners to offer for free.
- If the unsealing price is set to non-zero, retrievals will require the
  creation of a payment channel (even if price per byte is zero),
  which introduces significant latency and requires chain interaction.
```

proposals/lotus-dynamic-retrieval-pricing.md
```diff
@@ -0,0 +1,142 @@
...
+
+Impact of the blanket policy to price per byte:
+- Miners wanting to provide free retrieval for verified deals would set this
+  value to zero. This has the side-effect of skipping payment channel creation.
```
dirkmc, line 41:
```suggestion
  value to zero. This has the side-effect of skipping payment channel creation
  (if unseal price is also zero).
```

proposals/lotus-dynamic-retrieval-pricing.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### Alternatives
+_How might this project’s intent be realized in other ways (other than this project proposal)? What other potential solutions can address the same need?_
+
+The pricing function could be hardcoded, but that's too short-sighted.
+
+#### Dependencies/prerequisites
+<!--List any other projects that are dependencies/prerequisites for this project that is being pitched.-->
+
+F3 retrieval stabilisation.
```
dirkmc, line 127: I don't think this is really a pre-requisite - the ongoing stabilization work is for the case where the connection goes down in the middle of a data transfer, but a lot of retrieval deals will likely be made over stable connections.

raulk: Agree. I'm going to refine this to say that it's not technically a dependency, but work here needs to be coordinated with F3 as they are touching the same codebases.

raulk:
@dirkmc thanks for the review. I incorporated your coments.

dirkmc (approved):

proposals/lotus-miner-runtime-segregation.md
```diff
@@ -0,0 +1,99 @@
+# [outcome or objective here]
```
hannahhoward, line 1: should this be part of this proposal? It appears to lack content.

hannahhoward (approved):
General +1 to the utility of this project AND approach taken

proposals/lotus-dynamic-retrieval-pricing.md
```diff
@@ -0,0 +1,148 @@
...
+- If the unsealing price is set to non-zero, fast retrievals will require the
+  creation of a payment channel (even if price per byte is zero), which
+  introduces significant latency and requires chain interaction.
+
+Impact of the blanket policy to price per byte:
+- Miners wanting to provide free retrieval for verified deals would set this
+  value to zero. This has the side-effect of skipping payment channel creation 
+  (if unseal price is also zero).
+- However, that same policy would apply to unverified deals, which the miner is
+  likely not keen to offer for free.
```
ribasushi, line 48: Hm... this seems incongruent with what the goldenpath around retrieval seems to be?  That is: mining is so profitable that we do expect  in the short to medium term (multiple year) that miners will be offering free retrieval provided Fil+ is in place. Also dealbot efforts are centered around making unsealing untenable, by effectively requiring everything to be stored twice ( again - miners will go with it because it is cheaper to store anyway ).

Did thinking really shift that much since mid-march?

raulk: The goal hasn't changed. This line says that, currently, setting the retrieval price to byte to zero affects all storage deals, both verified AND unverified. That's one of the things we need to make dynamic.

ribasushi: I should have done a more precise highlight. Let me clarify:

- On unsealing: my understanding is that there is broad agreement ( including from CE ) that the act of unsealing is uneconomical given the current proofs system, and thus the option ought to be deprioritized if not deprecated within the "happy path". That is - miner always stores both sectors ( which is currently switched to be the default ) and if they do elect to delete some of the unsealed replicas, it is on them to pay the price for recovery. Additionally the dealbots ranking of TTFB will further make not providing fast retrieval untenable. Thus it seems odd at present expending effort and API-surface on making unsealing more visible.

- My other point was on how free and non-free retrieval are listed as equal options in the highlighted text, whereas it would be prudent to "nudge" both miner and client API users / implementors to gravitate towards "free by default, and not doing so is somewhat of a chore". In the spirit of "defaults matter" etc.

proposals/lotus-dynamic-retrieval-pricing.md
```diff
@@ -0,0 +1,148 @@
...
+
+![Solution sketch](./sketches/dynamic-retrieval-pricing.png)
+
+#### Brief plan of attack
+
+1. Introduce the ability to query the unsealing status of a piece from the
+   storage subsystem (unseal_status).
+2. Introduce the ability to query whether the storage deal was a verified deal
+   (verified_deal).
+3. Introduce the ability to query whether the storage deal had fast retrieval
```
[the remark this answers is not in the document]
aarshkshah1992: @hannahhoward 

IIUC, retrieval market creates dealIDs by using a counter. dealID is not a key that's shared between the storage & retrieval markets. 

During quoting a deal, the retrieval market does have access to the payloadCid and I am thinking of using that as a key to fetch the storage deal params we need for pricing. Would you know of a Lotus API that can help return the storage deal params given a payloadCid ?

The other solution I'm thinking of is injecting a storage market lookup function in the retreival markets that looks-up storage deals based on their payloadCid.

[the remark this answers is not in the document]
aarshkshah1992: @hannahhoward Ah, nvm. I think you mean the on chain dealID that the retrieval market can access using the PieceInfo. I see that Lotus has APIs to return storage params using the on chain dealID. This should be doable.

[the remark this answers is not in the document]
hannahhoward: @aarshkshah1992 yes your second comment is correct. I am referring to the on chain DealID. Eek terminology overloading!

jacobheun:
Released in Lotus 1.11.0, merging this.

PR #98, opened 2021-04-12
proposal: Lotus miner runtime segregation.

raulk:


kernelogic:
Similar to this https://github.com/filguard/lotus-ops/blob/master/documents/distributed-miner-configuration.md

raulk:
@kernelogic thank you for the pointer! Does that proposal segregate the markets logic from the sealing/mining logic in the way that is proposed here? Is the markets logic only running in daemon-public, or...?

hannahhoward (approved):
I will admit the existence of this project proposal warms my heart, cause go-fil-markets was always written with the goal of making this possible, which is part of why it felt important to maintain it as a seperate repo.

All interactions between go-fil-markets and the daemon / miner go through node interfaces:
https://github.com/filecoin-project/go-fil-markets/blob/master/storagemarket/nodes.go
https://github.com/filecoin-project/go-fil-markets/blob/master/retrievalmarket/nodes.go

Essentially if you can implement these by talking to daemons and miners over RPC, then you've seperated markets into its own process.

This isn't entirely simple -- the implementation of those interfaces are themselves somewhat complicated: https://github.com/filecoin-project/lotus/tree/master/markets/storageadapter & https://github.com/filecoin-project/lotus/tree/master/markets/retrievaladapter

Still hopefully this provides a pretty clear-ish path towards seperating out markets software. Yay that this is being considered!

raulk:
This is now released in https://github.com/filecoin-project/lotus/releases/tag/v1.11.1.

PR #103, opened 2021-05-20
Proposal: DeDHT - a DHT for Filecoin Deals

alanshaw:
Discovery of content providers on the Filecoin network without centralized indexes.

### DIFF
--- a/proposals/dedht.md
+++ b/proposals/dedht.md
@@ -1,0 +1,130 @@
+# DeDHT - a DHT for Filecoin Deals
+
+Authors: [@alanshaw](https://github.com/alanshaw)
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+<!--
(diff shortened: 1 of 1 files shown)


proposals/dedht.md
```diff
@@ -0,0 +1,130 @@
...
+
+On the surface Sybil attacks appear to be a problem. The more places in the DHT your miners are in the more deals they will receive. However if we mandate that participants in this DHT must be active miners registered in the Filecoin blockchain the barrier to entry to becoming a miner should be deterrent enough to prevent this.
+
+#### Assumptions &amp; hypotheses
+_What must be true for this project to matter?_
+<!--(bullet list)-->
+
+* Users want to discover which Filecoin miner is storing their data.
+* We don't really want to keep a central index of where the CIDs are stored.
+
```
vasco-santos, line 55: If I understood the approach correctly, we are also assuming that either the deal price does not vary significantly across miners or the deal price is not a problem for the user.

PR #104, opened 2021-05-25
proposal for pubsub boosters

vyzo:


raulk (approved):
This is a general engineering project (beyond product-market fit), and it can be delivered in one maintenance tribute. @vyzo and I are discussing executing on this for this week.

raulk (commented):
This also has the potential to eliminate https://github.com/filecoin-project/lotus/issues/5128, which has been a recurring complaint by miners that hasn't been fully addressed.

PR #105, opened 2021-05-25
Unixfsv1-directories-as-ADLs proposal

warpfork:
Long story short: this is a proposal to renovate the way we work with unixfsv1 data so that it becomes managable through the interface of an IPLD "ADL" (an "Advanced Data Layout"), which is an API concept design to handle needs like sharding.

The aim is that this should produce nicely maintainable code, while also fixing a lot of bugs along the way, providing a couple of nice features immediately (like Selectors working over unixfsv1 pathing, which is extremely neat), and providing a solid ground for new work.

### DIFF
--- a/proposals/unixfsv1-dir-as-adl.md
+++ b/proposals/unixfsv1-dir-as-adl.md
@@ -1,0 +1,384 @@
+Unixfsv1 directories as ADLs
+============================
+
+Authors: @warpfork, @lidel, and other reviewers
+
+Initial PR: TBD <!-- Reference the PR first proposing this document. Oooh, self-reference! -->
+
+Purpose & Impact
+----------------
+
+### Executive Summary
+
+Current unixfsv1 APIs provide bad developer experience in numerous ways,
+having direct bug-creation effects that are causing close collaborators to have serious problems,
+as well as having longer-term confusion effects which slow down development
+or at worst, may mislead development due to situations where the correct thing is not easy to do and a semantically mis-layered thing is easier than the correct thing.
+
+One example of a currently pressing problem is that users have stumbled upon a way of using the current APIs
+which results in the creation of data that is not actually transferable by libp2p (e.g. oversize blocks).
+While a series of "quick fixes" ("ducktape" solutions) to the most immediate problems are possible,
+there are still serious maintainability issues to address,
+and most of the "quick fixes" are user-hostile in some way (i.e. failing earlier in the path, and telling users to use a different API),
+or further complicated the developer experience in the long run (i.e. telling users to use a higher level API which avoids the most critical problem but provides _different semantics_ than the direct features the user wanted --
+which may further mislead any future development!).
+
+This proposal is for a targetted piece of development work which will solve this problems
+by supplying Unixfsv1 functionality through a clean implementation as an ADL (short for Advanced Data Layout,
+a concept from the IPLD project which is meant to provide a solution area for problems like sharding).
+
+In addition to solving the problem examples above,
+this work will also immediately result in several new features becoming reachable
+(such as IPLD Selectors becoming operable over Unixfsv1 pathing!,
+and build a strong foundation for future work.
+
+We believe that that the work outlined below will
+make the stack more robust,
+more consistent,
+and easier to understand for future developers both inside and nearby the core stack.
+
+### Background & intent
+_Describe the desired state of the world after this project? Why does that matter?_
+<!--
+Outline the status quo, including any relevant context on the problem you’re seeing that this project should solve. Wherever possible, include pains or problems that you’ve seen users experience to help motivate why solving this problem works towards top-line objectives.
+-->
+
+Unixfsv1 is a large and key subsystem of IPFS.
+It is the internal format used by IPFS whenever users are experiencing "files" and "directories".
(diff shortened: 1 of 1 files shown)


rvagg:
This needs to highlight right up front that this is strictly about the Go implementation--it's mentioned, but a little offhand and I don't think it'll be entirely obvious to many who are reading this.

Also IMO it'd be worth extra-highlighting that one of the highest points of leverage here is the independence of UnixFS and having it all in one, well-defined place with clean boundaries, and not having its logic spilling over in many places. Being able to use it independently and have it work for all of the difficult cases it needs to work (like scalability) out of the box is going to unlock a lot of potential ecosystem & independent tooling.

proposals/unixfsv1-dir-as-adl.md
```diff
@@ -0,0 +1,384 @@
...
+and the API boundaries are not especially clear.  Maintenance is generally difficult.
+It would be desirable for our own development and productivity if more of this implementation was unified in one place.
+
+At the same time, Unixfsv1's current implementation in golang predate much of the latest generation of work in IPLD.
+We believe significant gains of features, as well as clarity of design,
+are possible by rebuilding Unixfsv1 features in the form of an IPLD concept known as ADLs.
+
+Closer to users: The APIs made available to Unixfsv1 from IPFS can be confusing.
+Some very low-level features are made available by IPFS APIs, but are not always safe to use,
+because they may do incorrect things in "corner cases" which are not obvious to end users, resulting in a general air of fragility.
```
BigLep, line 61: Can we provide specific example(s)?

proposals/unixfsv1-dir-as-adl.md
```diff
@@ -0,0 +1,384 @@
...
+Some high level features are made available by other IPFS APIs, but jump to such a high level
+that users with specific needs can be observed to jump back down to the low level APIs
+(and then procede to use them incorrectly, because those APIs do not guide the user to good usage).
```
BigLep, line 64: Can we provide some example(s)?

proposals/unixfsv1-dir-as-adl.md
```diff
@@ -0,0 +1,384 @@
...
+
+Example user stories such as "patch this directory" currently don't work out well, because while we offer "patch" APIs in some places,
+they only work on low-level data, or, they work on high level data, but in ways that are not entirely desirable:
+
+- In the low-level way: we encounter problematic scenarios because low-level abstractions do not include enough knowledge
```
BigLep, line 69: Can we make clear what the low-level way is?  I assume it's `ipfs object patch add-link`

proposals/unixfsv1-dir-as-adl.md
```diff
@@ -0,0 +1,384 @@
...
+  to engage mechanisms like directory sharding... even when failure to do so means producing blocks of low-level data
+  which are so large that they exceed the limits set by data transfer mechanisms elsewhere in our stack.
+  This incoherence results in serious usability problems
+  (specifically, that users can create data which is then only available from that node, and cannot be transferred).
+- In the high-level way: While the "patch" user story can be satisfied (arguably) through the use of the MFS system,
```
BigLep, line 74: Make clear what this way is?  I assume it's `ipfs files cp`

BigLep (changes requested):
There's lots of good content here @warpfork and @lidel .  Thanks for putting it together.  

This isn't your fault, but I think the project proposal template may have gotten in the way of making the pitch more clear.  I worry that a non-engineer will get somewhat lost here.  I think we need to get *specific& problems/solutions from a customer's point-of-view called out upfront.  (I recognize a lot of time has been spent here and we don't want to spend a lot more.)

In reading through this again, I think a way (just an idea) we could crisply get the point across is with a structure like:

```
# Use Case #1: Customer wants to add a file to a directory
Incrementally adding a file to a directory is a common usecase for customers.

## Current Problematic Option #1: Use ipfs object patch
We have repeatedly seen customers reach for "patch" semantics.  It unfortunately has these issues:
1. no implicit gc protection - the result can be garbage collected before one can pin it
2. can create untransferrable block - there is no support for sharding of big directories.  A single block can be created that is too big to be exchanged over network

Because of these issues we have actually hidden the APIs and directed customers to use...

## Current Problematic Option #2: Use ipfs files cp
While this is our current recommended approach, it too is fraught with issues including:
1. Abstraction level
2. Built on top of the MFS library which few Stewards know how to touch, and those who do view it as radioactive
3. Clunky - have to type `files cp` and `files stat` 
...

## How use case is handled by ADLs
// Show what the user-typed code/calls look like when using ADLs

# Use Case #2: ...
... Repeated structure for use case above ...

# Other problems solved 
1. Unixfsv1 implementation components are currently scattered among many source code repositories,
and the API boundaries are not especially clear.  This makes maintenance and development difficult.  This approach would consolidate the implementation to a single repository.
2. Selectors
...
```

If this jives/makes sense, I would suggest putting content like this at the top and have the existing content below with a note that reader is welcome but not expected to read below.  I want a PM who commits 10 minutes of time to look at this to get why the engineering team would be recommending this.

As to how we prevent requiring so much write-up in the future, I'm looking to the improvements in https://github.com/protocol/web3-dev-team/issues/106 to help here. 

Thanks again for your time and work here.  I'm also game to talk more verbally.

OhArthits (approved):

Issue #106, opened 2021-06-01
Proposal: MPP (minimal project pitch) and adjustments to project proposal/prio process


pooja:
As discussed within the w3dt, how we operate in practice has evolved over the last few months (a significant change being the move to significant programs of work, as opposed to the individual-projects-first approach), motivating the desire to adjust our project proposal and prioritization process to match changes elsewhere in our operations.

## Process Changes

This diagram describes, at a high level, an iteration on "how we operate" that takes into consideration our current processes, and how we might desire for them to change going forward:
![project-process](https://user-images.githubusercontent.com/5668171/120250420-1032cd80-c24c-11eb-9d55-75a50be47034.png)

Some key changes to note:
1. This process introduces the idea of a **program scope document**, which synthesizes needs and information from various different sources to encapsulate a cohesive area of work, i.e. a "program." The program scope document lays out the motivation, goals, OKRs, KPIs, approach, and the set of projects that are being undertaken to achieve the program goals. This document would live within this GitHub repo.
2. This process introduces the idea of a **minimal project pitch (MPP)**. The MPP is a version of the [project proposal](https://github.com/protocol/web3-dev-team/blob/main/proposals/TEMPLATE.md) that is lighter weight since it inherits many of the proposal template information from the _program that it supports_. Below, we propose a revised set of fields to include in the MPP. It's also important to note that we recommend project teams write these MPPs after initial team alignment, brainstorming, and some initial program product reviews have happened. However, this is not required. It is very possible to propose a project for a program without having had detailed conversations about the project beforehand.
3. This process separates detailed technical design and project milestones/plan from the initial project pitch (MPP) so that we can agree that something _is generally a good idea to undertake and supports important goals_ before we invest heavily in designing _how_ we will undertake that project. 
4. This process creates room for projects that are not tied directly to programs as well.

Note also that w3dt teammates can always continue to use the original template if desired to provide more information about the project.

## Minimal Project Pitch (MPP) Template
Estimated to take <30minutes and be a couple of paragraphs long. Will be linked into the program scope document once accepted

- **Name of the project**
- **Program:** Does this project support a program? If so, which one?
- **Impact:**: What program goals/OKRs are being addressed -- why is this important? Or if this project is not proposed to be part of a program, why is this project important? What will we gain by picking up this project?
- **The idea:** Describe the proposed project solution. Diagrams and interface descriptions go a long way!
[...]

PR #110, opened 2021-06-14
Add MPP for filecoin.storage

pooja:


pooja:
Note that this project pitch follows the new [MPP template](https://github.com/protocol/web3-dev-team/blob/main/proposals/MINIMAL-PITCH-TEMPLATE.md) as per discussion in issue #106

alanshaw (approved):

olizilla (approved):

whyrusleeping:
It seems like the difference between this and estuary is a javascript client library. Would this project be just building and maintaining that library as a frontend to estuary? Otherwise i'm not entirely certain i understand the scope, or the scope seems to overlap very heavily with what estuary already offers, and it seems wasteful to duplicate work so much.

PR #111, opened 2021-06-15
Proposal: nft.storage Chunked CAR Uploads

alanshaw:
[Rendered](https://github.com/protocol/web3-dev-team/blob/7a2d2ec4931dcc250bceff26cd8a91d634665dfa/proposals/nft.storage-chunked-car-uploads.md)


proposals/nft.storage-chunked-car-uploads.md
```diff
@@ -0,0 +1,104 @@
...
+_Describe the status quo, including any relevant context on the problem you're seeing that this project should solve. Who is the user you're solving for, and why do they care about this problem? Wherever possible, include pain points or problems that you've seen users experience to help motivate why solving this problem works towards top-line objectives._ 
+
+NFT.storage has a hard limit of 100MB for uploads - a limit imposed by Cloudflare Workers (the server side technology that runs the site).
+
+This project proposes a way to work around this limit by sending _multiple_ requests of <100MB.
+
+## Impact
+_What goals/OKRs are being addressed (for w3dt, a specific program, etc.)? Why is this project important? What do we get with this project that we can't get without it?_
+
+Allows NFTs uploads of arbitrary size.
```
olizilla, line 15:
```suggestion
Allows NFT assets over 100MB to be reliably uploaded to nft.strorage.
```

proposals/nft.storage-chunked-car-uploads.md
```diff
@@ -0,0 +1,104 @@
...
+
+* An NFT is registered to the user before all data is received.
+* An out of order CAR chunk _may_ be garbage collected since it could be unreachable from the pinned root node.
+* Failure to upload all chunks will likely result in an NFT registered to a user with missing data forever.
+* A CAR file created with blocks in random order could get garbage collected.
+
+## Success/acceptance criteria (optional)
+_How do we know we're done with this project? How do we know we're successful? This field is OPTIONAL for the first draft of an MPP. Sometimes this field needs to be filled out once we have more detail on the shape of the actual solution._
+
+NFTs of >100MB can be uploaded to nft.storage.
```
olizilla, line 80:
```suggestion
Assets larger than 100MB can be uploaded to nft.storage.
```

olizilla (approved):
Nice! 

How about we set a goal for the size for asset uploads we will support... e.g up to 1GB? Technically we are just removing the aribitrary limit of the current code, so this could be any number, but I suspect we dont want unbounded uploads and having a target also gives us a clearer success criteria, and we can throw in stats for uploading 1Gb <or your target here> when it's shipped.

proposals/nft.storage-chunked-car-uploads.md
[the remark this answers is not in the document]
alanshaw: True, but I think maybe too much detail for this minimum project pitch (which is already pretty maximum).

proposals/nft.storage-chunked-car-uploads.md
```diff
@@ -0,0 +1,107 @@
...
+
+    This has the added benefit of allowing the creator to know the CID up front, allowing them to mint an NFT on a blockchain and upload the content to the nft.storage in parallel.
+
+    Note: the CAR file must be generated by the NFT creator using a tool such as [ipfs-car](https://github.com/vasco-santos/ipfs-car).
+    
+2. Split the CAR file into chunks of <100MB.
+
+    The splitting process does not attempt to understand the DAG, it simply reads the blocks until it has reached 100MB and then starts creating a new CAR with the remaining blocks.
+    
+    Each CAR has a "chunk root" node which references all the blocks and roots (only the first chunk) in the CAR.
```
aschmahmann, line 32: Is the chunk root listed as a root in the CAR file? It seems like there are some (very solvable) edge cases here around the maximum size that the chunker would need to be aware of. Some examples include:

1) If there are lots of blocks that need to go under the root needing to build a HAMT for the chunk root. Not counting the size of the chunk root related data could push you over the limit.
2) If there are large identity CIDs then you could run into trouble, particularly with the chunk root because you'll be duplicating the data.

mikeal:
I thought we swapped the ordering and the root pin happens first?

proposals/nft.storage-chunked-car-uploads.md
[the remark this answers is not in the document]
mikeal: We can get around this later by doing an encoding step where we throw away the raw blocks and send the top of the graph first, then we can do another encode, stream the raw blocks to CAR files under 1GB and await continued encoding until the write for each CAR completes as we create them. That would work as long as available memory is greater than 1GB.

proposals/nft.storage-chunked-car-uploads.md
```diff
@@ -0,0 +1,107 @@
...
+* An out of order CAR chunk _may_ be garbage collected since it could be unreachable from the pinned root node.
+* Failure to upload all chunks will likely result in an NFT registered to a user with missing data forever.
+* A CAR file created with blocks in random order could get garbage collected.
+
+## Success/acceptance criteria (optional)
+_How do we know we're done with this project? How do we know we're successful? This field is OPTIONAL for the first draft of an MPP. Sometimes this field needs to be filled out once we have more detail on the shape of the actual solution._
+
+* Assets larger than 100MB can be uploaded to nft.storage.
+* A user uploads a 1GB asset to nft.storage.
+
```
lidel, line 84:
```suggestion
* Upload endpoint is part of [ipfs/pinning-services-api-spec](https://github.com/ipfs/pinning-services-api-spec/)

```
Could this proposal be an opportunity to come up with a spec for a generic endpoint that any public pinning service could provide? Feels like something we discussed in https://github.com/ipfs/pinning-services-api-spec/issues/7 

My main drive here is to remove vendor lock-in and make all the CAR-splitting tooling usable outside nft.storage (removing barriers if someone chooses to run their own infra).

I was tempted to add "ipfs-cluster  can be used for self-hosting", but I am not sure if nft.storage uses it directly. Perhaps ["making ipfs-cluster compatible with pinning service api"](https://github.com/ipfs/ipfs-cluster/issues/1213) should be a separate project on its own?

aschmahmann: I think you meant ipfs/pinning-services-api-spec#73, right?

> My main drive here is to remove vendor lock-in and make all the CAR-splitting tooling usable outside nft.storage (removing barriers if someone chooses to run their own infra).

Do you mean supporting single car file uploads via the pinning service API, or this multi car file approach? Note that the atomicity properties which were the subject of some debate in that issue are not really present in this multi-car approach which makes things pretty easy (although we could still push on the atomic upload + pin of a single car file). On the other hand, the multi-car approach taken here seems pretty specific to some assumptions about nft.storage which makes asking other people to implement it seem like a stretch.

IIUC some assumptions in this proposal which are coupled to nft.storage are:
- The expiry time ensures that the chunk root nodes can eventually be garbage collected and also ensures that uploads that never complete can eventually be garbage collected.
    - Implies the application can properly track + label pins for the same chunk root even by the same user (e.g. uploading overlapping data twice)
    - Implies the application can support time-expired pins and garbage collection of them
- Pinning the root last also ensures nft.storage has all the data for the NFT before it is registered to the user. It's an insentive to continue uploading all the chunks, avoiding any blame for losing data in the case where a user ceases uploading before completion.
     - Implies the application is unable/unwilling to inform the user of whether their upload is in progress or not (e.g. when I upload a 50MB file to GitHub I can see the progress report of how far along it is)

proposals/nft.storage-chunked-car-uploads.md
aschmahmann: Side note: If we were willing to pin the roots first then the complexity around the edge cases would go away and you could literally just grab 100MB worth of blocks (minus the header size).

proposals/nft.storage-chunked-car-uploads.md
alanshaw: I'll reply to these comments properly tomorrow but just quickly  - IPFS Cluster provides the expiring pin functionality.

olizilla:
> Could this proposal be an opportunity to come up with a spec for a generic endpoint

The implementation details of this proposal are tied to the infra that nft.storage is using. We're not looking to solve more general cases yet. This work just provides a data point for those conversations.

> I thought we swapped the ordering and the root pin happens first?

That was the initial plan, and is described in the [Alternative section](https://github.com/protocol/web3-dev-team/blob/ff76251e76eb0d12c900bca44d3f71cdeb2b02f1/proposals/nft.storage-chunked-car-uploads.md#alternative). As cluster support `/add` with an expiration date, the current proposal is to send the other chunks first, and have cluster handle removing them if the root never arrives. I think this means the logic for splitting the CAR is simpler, and Alan's `carbites` module can be used as is.

vasco-santos:
> I thought we swapped the ordering and the root pin happens first?

Complementing what @olizilla mentioned, we also do not end up with inconsistent state in the KV, as with that approach dropped uploads would have the NFT registered (given it would be registered in the first request)

alanshaw:
Just to follow up on this - we actually ended up going with a different strategy that splits CARs so they all have a same root CID but contain different sub-graphs. It means there's no special "chunk root" node and no need for pin expiry. The trade off is having to understand (decode blocks) and traverse the DAG (as well as some other minor trades 😉). Its closer to the "alternative" mentioned in the PR.

[Carbites](https://github.com/alanshaw/carbites#usage) now supports multiple strategies so you can pick your trade offs.

I'll try to update the PR to reflect the route we took when I next have a moment 😬

PR #120, opened 2021-06-25
proposal for extended blockstore interface

vyzo:


proposals/lotus-blockstore-interface.md
```diff
@@ -0,0 +1,44 @@
...
+type Blockstore interface {
+  ...
+  Get(c cid.Cid, opts Options) (blocks.Block, error)
+  ...
+}
+```
+
+We may want to name the new option accepting methods with an `WithOptions` suffix if we want to keep backwards compatibility in the interface.
+
+Note: The reason for using a flat `Options` struct, as opposed to a `context.Context` object or functional options is purely performance. The latter options can have quite a bit of overhead, which we want to eliminate.
```
aschmahmann, line 34: We want to put contexts everywhere anyway for the purposes of debugging/tracing as well as general extensibility https://github.com/ipfs/go-ipfs/issues/6803.

Is the performance of `context.WithValue` really that slow? If so then I could see options making sense here.

proposals/lotus-blockstore-interface.md
```diff
@@ -0,0 +1,44 @@
...
+
+## The idea
+
+Extend the blockstore interface so as to accept an `Options` struct, which specifies the necessary context constraints.
+
+Methods would look like this:
+```
+type Blockstore interface {
+  ...
+  Get(c cid.Cid, opts Options) (blocks.Block, error)
```
aschmahmann, line 27: Why `opts Options` instead of `opts Options...`? Are there only options for Get, what about the other operations on the blockstore?

proposals/lotus-blockstore-interface.md
```diff
@@ -0,0 +1,44 @@
...
+This is both a performance/gc issue (the hotstore is larger than it needs to be) but it is also a potential attack vector for nodes that need to run with fixed hardware (e.g. network boosters).
+- We lack the ability to inform the splitstore that certain objects should be reified from cold storage to hot storage. For instance, a miner doing tipset state computation in an older tipset may want to keep this in the hot storage. It also means that misses result in permanent banishment in the coldstore and makes it hard to gc by simply removing the coldstore.
+
+## Impact
+
+This will make the splitstore a much more reliable component, working towards making it the default blockstore in lotus.
+
+## The idea
+
+Extend the blockstore interface so as to accept an `Options` struct, which specifies the necessary context constraints.
```
aschmahmann, line 21: Is this only within lotus or are you proposing to modify https://github.com/ipfs/go-ipfs-blockstore?

proposals/lotus-blockstore-interface.md
vyzo: it can be quite bad in my experience, especially with multiple options and compound contexts.

proposals/lotus-blockstore-interface.md
vyzo: so that we don't have to parse basically.

proposals/lotus-blockstore-interface.md
vyzo: This is for the lotus blockstore, changing the IPFS blockstore is harder.

proposals/lotus-blockstore-interface.md
vyzo: Also, it's for all methods, not just Get -- I just didnt want to list them all here.

And and and, I really want to kill the `AllKeysChan` atrocity and replace it with `ForEach`.

proposals/lotus-blockstore-interface.md
```diff
@@ -0,0 +1,44 @@
+# Extended Blockstore Interface
+
+Authors: vyzo
+
+## What is the problem this project solves?
+
+We currently have a limited "dumb" blockstore interface, which does not allow us to pass context options so as to inform the blockstore about certain properties of objects being stored.
+
+This is particularly important in the context of the splitstore, where we have the following problems:
+- We lack the ability to inform the splitstore of the actual epoch an object belongs, which requires us to "guess" by estimating based int he current tipset and wall clock (see [#6474](https://github.com/filecoin-project/lotus/pull/6474)). This is problematic for many reasons, as an object might be misassigned its epoch (and its completely broken during catch up sync in current master).
```
raulk, line 10:
```suggestion
- We lack the ability to inform the splitstore of the actual epoch an object belongs, which requires us to "guess" by estimating based on the current tipset and wall clock (see [#6474](https://github.com/filecoin-project/lotus/pull/6474)). This is problematic for many reasons, as an object might be misassigned its epoch (and thus become completely broken during catch-up sync in current master).
```

proposals/lotus-blockstore-interface.md
```diff
@@ -0,0 +1,44 @@
...
+- We lack the ability to inform the splitstore when an access is initiated from the network. In conjunction with the transactional gc logic in [#6474](https://github.com/filecoin-project/lotus/pull/6474) this would result in network requests keeping objects that would otherwise be pruged from the hotstore live.
```
raulk, line 11:
```suggestion
- We lack the ability to inform the splitstore when an access is initiated from the network. In conjunction with the transactional gc logic in [#6474](https://github.com/filecoin-project/lotus/pull/6474) this would result in network requests keeping alive objects that would otherwise be purged from the hotstore.
```

raulk: The `ExposedBlockstore` abstraction is not sufficient.

proposals/lotus-blockstore-interface.md
```diff
@@ -0,0 +1,44 @@
...
+This is both a performance/gc issue (the hotstore is larger than it needs to be) but it is also a potential attack vector for nodes that need to run with fixed hardware (e.g. network boosters).
+- We lack the ability to inform the splitstore that certain objects should be reified from cold storage to hot storage. For instance, a miner doing tipset state computation in an older tipset may want to keep this in the hot storage. It also means that misses result in permanent banishment in the coldstore and makes it hard to gc by simply removing the coldstore.
```
raulk, line 13: I don't think this should be possible. This does not interact well with configurable retention policies nor with GC. If the node operator configures a 10000 epoch retention, it should just get 10000 epochs. Doing this is too much voodoo, as suddenly you have no idea of what's actually in the hotstore. I think there's value in defining the hotstore as **strictly what the VM needs to operate correctly**, and not that + a random collection of objects that a lucky user may have managed to page back in during a magic time window.

Also: allowing this flow opens up an attack vector where an attacker could just page everything in just before the deadline. They keep doing this ad infinitum and suddenly the hot store _is_ the cold store.

proposals/lotus-blockstore-interface.md
raulk: This is a low-level operation, we can't afford to incur in the cost of heap-escaping functional options for every invocation.

proposals/lotus-blockstore-interface.md
raulk: `context.WithValue` is also not typed.

Stebalien: There's no way we can both make `Options` "typed" _and_ support passing things like epochs (which I don't think we should do anyways).

proposals/lotus-blockstore-interface.md
```diff
@@ -0,0 +1,44 @@
+# Extended Blockstore Interface
+
+Authors: vyzo
+
+## What is the problem this project solves?
+
+We currently have a limited "dumb" blockstore interface, which does not allow us to pass context options so as to inform the blockstore about certain properties of objects being stored.
+
+This is particularly important in the context of the splitstore, where we have the following problems:
+- We lack the ability to inform the splitstore of the actual epoch an object belongs, which requires us to "guess" by estimating based on the current tipset and wall clock (see [#6474](https://github.com/filecoin-project/lotus/pull/6474)). This is problematic for many reasons, as an object might be misassigned its epoch (and thus become completely broken during catch-up sync in current master).
```
Stebalien, line 10: This is very much the wrong way to go about this. Instead, the blockstore should be _managed_ by a system on-top of the blockstore that understands what we need to keep and what we don't need to keep.

It sounds like we need some form of tagged and/or generational blockstore:

1. Low-level blockstore that stores only blocks.
2. Intermediate service with a `ForEpoch(epoch) Blockstore` function that returns a blockstore for _that_ epoch.

Stebalien (commented):
This proposal mixes concerns. The blockstore is concerned with storing opaque blocks, not with pinning, GC, etc.

proposals/lotus-blockstore-interface.md
iand: I think this proposal would be a lot stronger with a concrete list of options that would apply to a Get operation. Currently it appears to be a catch-all for any possible future behavior.

I'm also moderately perturbed by having per-operation options rather than per-store. This implies a potential large variation in performance characteristics per operation.

proposals/lotus-blockstore-interface.md
aschmahmann: If we're breaking the interfaces anyway any reason not to add in a context? Seems like it'd be helpful if we land the changes to the blockstore + datastore linked above which we're hoping to do.

vyzo:
Closing as the splitstore has been redesigned to obviate the need for radical interface changes.

PR #121, opened 2021-06-25
prposal for formal lookback limits

vyzo:


Stebalien:
1. If we actually need to set limits, this would need to be a FIP and would need to be enforced in actors. However, I think we just need to figure out what the existing limits are.
2. In terms of _state_ lookback: We lookback one chain finality (900 epochs) for the block miner's worker key & power. I'm not aware of _any_ other lookback.
3. In terms of _chain_ lookback (for randomness/beacons), the answer is likely a month plus a bit. But _really_, there's no reason not to keep the entire chain (not the messages, just the block headers).

So the answer is: 900 epochs (1800-2000 to be safe) plus the chain headers.

Stebalien:
Anyways, what is a "booster". Are you referring to public nodes for light clients? _Those_ nodes can probably reject all requests that try to access anything not in the current tipset. They can also _refuse_ to execute any slashing messages not included in a block.

